PR #1, opened 2023-01-30
SSE client, example, and README

jspahrsummers:


src/index.ts
```diff
@@ -6,3 +8,88 @@ export type StreamingMessage = {
...
+        },
+        onmessage: (ev) => {
+          if (ev.event === Event.Ping) {
+            return;
+          }
+
+          const message = JSON.parse(ev.data) as StreamingMessage;
+          if (message.stop_reason !== null) {
+            abortController.abort();
+            return resolve(message.completion);
```
mikelambert, line 75: Is there a way for the caller to see why it stopped? (ie, did I hit max_tokens, or did I hit the stop token I specified?)

src/index.ts
```diff
@@ -6,3 +8,88 @@ export type StreamingMessage = {
...
+  stop_sequences: string[];
+  max_tokens_to_sample: number;
+};
+
+export type OnSampleChange = (sample: string) => void | Promise<void>;
+
+export const HUMAN_PROMPT = "\n\nHuman:";
+export const AI_PROMPT = "\n\nAssistant:";
+
+const CLIENT_ID = "anthropic-typescript/0.2.0";
```
mikelambert, line 15: Can this be dynamically pulled from your package.json version? I can easily see this falling out of sync.

src/index.ts
```diff
@@ -6,3 +8,88 @@ export type StreamingMessage = {
...
+}
+
+export class Client {
+  private apiUrl: string;
+
+  constructor(private apiKey: string, options?: { apiUrl?: string }) {
+    this.apiUrl = options?.apiUrl ?? DEFAULT_API_URL;
+  }
+
+  sample(
```
mikelambert, line 38: Given there are both stream (get every intermediate sample) and non-stream (wait until you have the full result, then give it to me) versions of the API, should we be calling this one a suffix of "_stream" or something? (I did this in the python one)

Also, should this be called "complete" or "complete_stream" to match the underlying openapi.json/REST endpoint name, and make it easier for someone to understand/lookup? (Another thing I changed on the python SDK)

mikelambert (approved):
On

src/index.ts
jspahrsummers: I looked around, and the solutions I found for npm packages are quite hacky. Most involve importing `package.json` (bad, because then it bundles the full file with the source) or some sort of build-time generation of a dynamic file + import.

I agree with you, but not sure the alternatives are better at the moment

src/index.ts
jspahrsummers: > Given there are both stream (get every intermediate sample) and non-stream (wait until you have the full result, then give it to me) versions of the API, should we be calling this one a suffix of "_stream" or something?

I didn't write the support for this yet, but can do.

> Also, should this be called "complete" or "complete_stream" to match the underlying openapi.json/REST endpoint name, and make it easier for someone to understand/lookup? (Another thing I changed on the python SDK)

👍

src/index.ts
jspahrsummers: Addressed in #2

src/index.ts
jspahrsummers: Addressed in #2

jspahrsummers:
@mikelambert Thanks for the review! My auto-merge was a bit too aggressive, so I've addressed your comments in #2

PR #5, opened 2023-02-02
Add optional `onOpen` callback to `completeStream`

jspahrsummers:
This can be used to tell when the stream has first begun, instead of having to wait for the first token.

# To do

## After merging

- [x] Release

### DIFF
--- a/examples/basic_stream.ts
+++ b/examples/basic_stream.ts
@@ -16,7 +16,14 @@ client
       max_tokens_to_sample: 200,
       model: "claude-v1",
     },
-    (updatedSample) => console.log(updatedSample.completion)
+    {
+      onOpen: (response) => {
+        console.log("Opened stream, HTTP status code", response.status);
+      },
+      onSampleChange: (updatedSample) => {
+        console.log(updatedSample.completion);
+      },
+    }
   )
   .then((finalSample) => {
     console.log("Finished sampling:\n", finalSample.completion);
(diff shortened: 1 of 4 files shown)


examples/basic_stream.ts
```diff
@@ -16,7 +16,14 @@ client
       max_tokens_to_sample: 200,
       model: "claude-v1",
     },
-    (updatedSample) => console.log(updatedSample.completion)
+    {
+      onOpen: (response) => {
+        console.log("Opened stream, HTTP status code", response.status);
+      },
+      onSampleChange: (updatedSample) => {
```
jacksonkernion: No need to change this, but since the api endpoint is now `complete`, `onSampleChange` is 
 bit confusing. Maybe change to `onCompletionStep`? Not sure if anyone's actually using this yet...

jacksonkernion (approved):
Nice addition 👍

examples/basic_stream.ts
jspahrsummers: It's okay, this PR is a breaking change anyway (because the callbacks are now in an object instead of positional parameters).

I don't mind changing this callback, but I feel "change" or "update" or "sample" is important verbiage to have in there somewhere, and it might be confusing to call it a "completion" (even though it is!) when the API call has not yet completed.

What do you think?

jacksonkernion: I think we can treat 'completion' as a technical term for LLMs--our developers should know that since the endpoint is called 'complete'. I find the inconsistency between sample and completion more jarring myself. (There's nothing called 'sample' in the API request/response.)

But I'm not wedded to 'completion'. I like 'update' over 'change'. Maybe just `onUpdate`?

PR #18, opened 2023-05-11
Make SDK throw actual JS Error instances on errors

jacoblee93:
There are a few instances where the SDK rejects with normal objects instead of instances of the JS `Error` class, which results in confusing issues when using the SDK downstream. I've updated that here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@anthropic-ai/sdk",
-  "version": "0.4.3",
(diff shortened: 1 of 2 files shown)


jspahrsummers (approved):
Great—thanks for the fix!

rattrayalex:
We've rewritten the SDK, and proper Error instances should be thrown in all cases. Thank you for this!

PR #20, opened 2023-06-01
Add onClose() callback

jaredpalmer:
Add an `onClose()` callback. This is very useful for folks who want to dually stream and accumulate the response (for saving to a DB).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,7 @@ export type SamplingParameters = {
 
 export type OnOpen = (response: Response) => void | Promise<void>;
 export type OnUpdate = (completion: CompletionResponse) => void | Promise<void>;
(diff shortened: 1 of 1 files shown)


jspahrsummers:
The intention right now is that you can chain off of the returned `Promise` to accomplish things like this. Would that work for you?

rattrayalex:
Thanks @jaredpalmer ! The API has been overhauled, so this PR is no longer relevant. 

I believe you can now accomplish this with `stream.controller.on('abort', myCallback)` if needed, though the following should work for most use-cases:

```ts
let text;
for await (const part of await client.completions.create({...params, stream: true})) {
  text += part.completion
  mystream.push(part) // or equivalent
}
await db.save(text)
```

If neither of these work well for your use-case, please let us know; feedback very welcome.

PR #77, opened 2023-07-19
feat(config): export `Config` type

arielweinberger:
We're building an Anthropic integration for [Pezzo](https://pezzo.ai) and need to access the Anthropic `Config` object.

Since it isn't exported at the moment, we find ourselves having to duplicate the type, which is obviously less than ideal due to future deviations and having to keep up with the up-to-date type definition.

This PR exports it.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,7 +7,7 @@ import * as Errors from './error';
 import type { Agent } from '@anthropic-ai/sdk/_shims/agent';
 import * as Uploads from './uploads';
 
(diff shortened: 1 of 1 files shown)


rattrayalex:
Ah, sorry about this oversight and thank you for the contribution! We actually just added this in https://github.com/anthropics/anthropic-sdk-typescript/pull/75 – it simply hasn't landed in `main` yet. We'll do that shortly and get this out to you!

PR #85, opened 2023-07-27
fix 'process is not defined' when not providing authToken

implicit-invocation:


rattrayalex:
Thank you for the contribution! We've refactored this a bit differently, which has landed in https://github.com/anthropics/anthropic-sdk-typescript/pull/90 and should resolve the issue.

PR #242, opened 2023-12-20
Fix header shadowing for edge functions

jacoblee93:
The newest SDK version seems to have added an issue where two variants of the `x-api-key` header get set:

```
{
      'Content-Length': '175',
      Accept: 'application/json',
      'Content-Type': 'application/json',
      'User-Agent': 'Anthropic/JS 0.11.0',
      'X-Stainless-Lang': 'js',
      'X-Stainless-Package-Version': '0.11.0',
      'X-Stainless-OS': 'MacOS',
      'X-Stainless-Arch': 'arm64',
      'X-Stainless-Runtime': 'node',
      'X-Stainless-Runtime-Version': 'v18.18.2',
      'X-Api-Key': '<secret key here>',
      'anthropic-version': '2023-06-01',
      'Anthropic-Beta': 'messages-2023-12-15',
      'anthropic-beta': 'messages-2023-12-15',
      'x-api-key': ''
}
```

This causes undefined behavior - it seems Node will dedupe using `X-Api-Key` while edge runtimes will use the lower cased version, resulting in mysterious auth issues.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -136,10 +136,10 @@ export class Anthropic extends Core.APIClient {
   }
 
   protected override validateHeaders(headers: Core.Headers, customHeaders: Core.Headers) {
-    if (this.apiKey && headers['X-Api-Key']) {
+    if (this.apiKey && headers['x-api-key']) {
       return;
     }
-    if (customHeaders['X-Api-Key'] === null) {
+    if (customHeaders['x-api-key'] === null) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


jacoblee93:
CC @rattrayalex

rattrayalex:
Thank you very much @jacoblee93 ! We've fixed this in https://github.com/anthropics/anthropic-sdk-typescript/releases/tag/v0.12.0

jacoblee93:
Awesome! Will bump

PR #320, opened 2024-03-06
--save hasn't been needed since 2017

spullara:
The requirement was removed in the NPM 5.0.0 release.


rattrayalex (approved):
Thanks for catching that & contributing!

PR #358, opened 2024-03-26
fix: use anthropic-ratelimit headers

### LINKED ISSUE #357: retry-after header missing from response
If there's no `retry-after` or `retry-after-ms` response header, the library should create one from the greater value of the `anthropic-ratelimit-requests-reset` and `anthropic-ratelimit-tokens-reset` headers (if both `anthropic-ratelimit-requests-remaining` and `anthropic-ratelimit-tokens-remaining` are zero). If only one of those headers is zero, use the reset time for that one.

```
{
    'anthropic-ratelimit-requests-limit': '5',
    'anthropic-ratelimit-requests-remaining': '0',
    'anthropic-ratelimit-requests-reset': '2024-03-26T20:00:00Z',
    'anthropic-ratelimit-tokens-limit': '25000',
    'anthropic-ratelimit-tokens-remaining': '16000',
    'anthropic-ratelimit-tokens-reset': '2024-03-26T20:00:00Z',
    'cf-cache-status': 'DYNAMIC',
    'cf-ray': '86a9b96cdac90359-MIA',
    connection: 'keep-alive',
    'content-length': '305',
    'content-type': 'application/json',
    date: 'Tue, 26 Mar 2024 19:59:59 GMT',
    'request-id': 'req_01T5fCVnYdf9X5ZW6KayReBK',
    server: 'cloudflare',
    via: '1.1 google',
    'x-cloud-trace-context': 'ae6af598607da922ea75fbb21d9c3bd6',
    'x-should-retry': 'true'
  }
```

aleclarson:
Closes #357

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -548,23 +548,48 @@ export abstract class APIClient {
   ): Promise<APIResponseProps> {
     let timeoutMillis: number | undefined;
 
-    // Note the `retry-after-ms` header may not be standard, but is a good idea and we'd like proactive support for it.
-    const retryAfterMillisHeader = responseHeaders?.['retry-after-ms'];
-    if (retryAfterMillisHeader) {
-      const timeoutMs = parseFloat(retryAfterMillisHeader);
-      if (!Number.isNaN(timeoutMs)) {
-        timeoutMillis = timeoutMs;
+    if (responseHeaders) {
+      // Note the `retry-after-ms` header may not be standard, but is a good idea and we'd like proactive support for it.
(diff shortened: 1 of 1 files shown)


rattrayalex (commented):
Thanks for this! The core of what you have here looks plausible to me, though we'd want to tweak it to be more additive (since it's based on codegen). 

As mentioned in the linked issue, let's hold for now to see if we can just change backend behavior.

rattrayalex:
Anthropic now always sends `retry-after` headers in these scenarios, so this change is no longer needed.

PR #362, opened 2024-03-29
fix: update `output_tokens` when streaming

mdjastrzebski:
## Summary

When using streaming API (`antrophic.messages.stream`), the final message (`stream.finalMessage()`) always reports single output token:

```
    // ...
    "model": "claude-3-sonnet-20240229",
    "stop_reason": "end_turn",
    "stop_sequence": null,
    "usage": {
      "input_tokens": 199,
      "output_tokens": 1
    }
```

## Fix 

On `message_delta` event update the `snapshot.usage.output_tokens` field from event output tokens.

### DIFF
--- a/src/lib/MessageStream.ts
+++ b/src/lib/MessageStream.ts
@@ -450,6 +450,7 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
       case 'message_delta':
         snapshot.stop_reason = event.delta.stop_reason;
         snapshot.stop_sequence = event.delta.stop_sequence;
+        snapshot.usage.output_tokens = event.usage.output_tokens;
         return snapshot;
       case 'content_block_start':
         snapshot.content.push(event.content_block);


rattrayalex:
This looks like a port of https://github.com/anthropics/anthropic-sdk-python/pull/426 – we may already have it in-flight on our end, otherwise this should be good to merge.

Regardless, thanks for the fix!

mdjastrzebski:
I wasn't aware of Python issue/fix. I've encountered this when testing my app and figured out that should be a proper fix 😄

rattrayalex:
It looks like we've already made this fix, it's just pending release: https://github.com/anthropics/anthropic-sdk-typescript/pull/348/files#diff-a3e76193753c78182d693458d0af64582c15fcc356dd4668ab954787392c195fR453

mdjastrzebski:
Closing, as the issue seems to be fixed in SDK v0.19.1.

PR #366, opened 2024-03-31
Proposed changed to update punycode  

### LINKED ISSUE #365: punycode update?
```
  (node:16104) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.                                                                                         (Use `node --trace-deprecation ...` to show where the warning was created) 
  ```
  I get this error when i try to run a lua file after configuring node js to load a js file and run directly in the lua file and after testing it with basic log out it worked but when i decide to use it and run the main thing which is my js application it faile

### PR DESCRIPTION
**Pull request by urizennnn**
I just did yarn add punycode and changed dependencies that require punycode to fix #365

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "form-data-encoder": "1.7.2",
     "formdata-node": "^4.3.2",
     "node-fetch": "^2.6.7",
+    "punycode": "^2.3.1",
     "web-streams-polyfill": "^3.2.1"
   },
   "devDependencies": {
--- a/packages/bedrock-sdk/yarn.lock
+++ b/packages/bedrock-sdk/yarn.lock
@@ -4247,7 +4247,7 @@ uri-js@^4.2.2:
   resolved "https://registry.yarnpkg.com/uri-js/-/uri-js-4.4.1.tgz#9b1a52595225859e55f669d928f88c6c57f2a77e"
   integrity sha512-7rKUyy33Q1yc98pQ1DAmLtwX109F7TIfWlW1Ydo8Wl1ii1SeHieeh0HHfPeL2fMXK6z0s8ecKs9frCuLJvndBg==
   dependencies:
-    punycode "^2.1.0"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by dgellow** [comment]
Hi @urizennnn. Unfortunately I do not think that would be enough. The problem here is that `require("punycode")` imports the punycode node core library that has been deprecated. `whatwg-url`, a dependency of `node-fetch` does that import. Even if we add `punycode` explicitly the core library would still be the one being imported.

To get rid of the warning we would either need to replace `node-fetch` (the work would be similar to what is done with https://github.com/openai/openai-node/pull/402), or have users of the sdk patch `require("punycode")` in whatwg-url to become `require("punycode/")` (using a patch utility such as https://github.com/ds300/patch-package?tab=readme-ov-file).

**Comment by urizennnn** [comment]
> Hi @urizennnn. Unfortunately I do not think that would be enough. The problem here is that `require("punycode")` imports the punycode node core library that has been deprecated. `whatwg-url`, a dependency of `node-fetch` does that import. Even if we add `punycode` explicitly the core library would still be the one being imported.
> 
> To get rid of the warning we would either need to replace `node-fetch` (the work would be similar to what is done with [openai/openai-node#402](https://github.com/openai/openai-node/pull/402)), or patch `require("punycode")` in whatwg-url to become `require("punycode/")`.

oh ok, so are there any temporal fixes available atm?

**Comment by rattrayalex** [comment]
In the short term, you can work around this by overriding the dependency in your package.json, eg:

```json
"overrides": {
  "whatwg-url": "13.0.0"
}
```

urizennnn:
> In the short term, you can work around this by overriding the dependency in your package.json, eg:
> 
> ```json
> "overrides": {
>   "whatwg-url": "13.0.0"
> }
> ```

Considering my case that won't work since I am not running the code in native js (not sure that is the right way to explain but yh), it being ran from a lua file so i'm not sure it would fix that but i just added the ``` --no-deprecation``` flag to it

Issue #380, opened 2024-04-06
Bedrock-sdk: Vercel edge runtime & Cloudflare not supported


florianjuengermann:
When using the `@anthropic-ai/bedrock-sdk` in nextjs/vercel edge functions, it fails with 
```
Attempted import error: 'fromNodeProviderChain' is not exported from '@aws-sdk/credential-providers' (imported as 'fromNodeProviderChain').

Import trace for requested module:
./node_modules/.pnpm/@anthropic-ai+bedrock-sdk@0.9.5_encoding@0.1.13/node_modules/@anthropic-ai/bedrock-sdk/auth.mjs
```

According to the [anthropic bedrock-sdk readme](https://github.com/anthropics/anthropic-sdk-typescript/tree/main/packages/bedrock-sdk), Vercel edge functions are supported.

But in [bedrock-sdk/src/auth.ts#L3](https://github.com/anthropics/anthropic-sdk-typescript/blob/bbce9d053b68c55b4b5961f5fe7cbb177838d18c/packages/bedrock-sdk/src/auth.ts#L3C49-L3C69) `fromNodeProviderChain` is imported from `@aws-sdk/credential-providers` which according to [their readme](https://www.npmjs.com/package/@aws-sdk/credential-providers#fromNodeProviderChain), is "Not available in browsers & native apps" and therefore probably not available in the edge runtime.

Are there any trick to get it to run in the edge runtime?

rattrayalex:
Thanks for raising this, we'll take a look soon.

prajwalpkn:
Any updates on this?

rattrayalex:
Sorry we let this slip. I actually think that with AWS's auth system, we won't be able to support bedrock on edge runtimes, but we'll take a look soon and let you know.

domleboss97:
Given that this seems to be an issue on edge runtimes, it might be good to, at the very least, update the [readme](https://github.com/anthropics/anthropic-sdk-typescript/tree/main/packages/bedrock-sdk) to reflect reality as it relates to supported environments. CF workers and vercel edge are both listed there.

mitsuhiko:
If someone else is running into this: the same issue exists if you try to to the same from CloudFlare workers.

rattrayalex:
Thanks for the bump Armin. Not great for this bug to have gone unresolved for so long, I'm sorry about that.

RobertCraigie:
[...]

Issue #314, opened 2024-03-06
Could not resolve a few of packages in @anthropic-ai/bedrock-sdk


Code-Hex:
Environment:
- wrangler@3.31.0
- miniflare@3.20240223.1
- typescript@5.3.3

Enabled `compatibility_flags = [ "nodejs_compat" ]` in `wrangler.toml`
https://developers.cloudflare.com/workers/runtime-apis/nodejs/#enable-nodejs-with-workers

```
✘ [ERROR] Could not resolve "assert"

    ../../node_modules/.pnpm/@anthropic-ai+bedrock-sdk@0.9.0/node_modules/@anthropic-ai/bedrock-sdk/auth.mjs:1:19:
      1 │ import assert from 'assert';
        ╵                    ~~~~~~~~

  The package "assert" wasn't found on the file system but is built into node.
  Add "node_compat = true" to your wrangler.toml file and make sure to prefix the module name with "node:" to enable Node.js compatibility.


✘ [ERROR] Could not resolve "stream"

    ../../node_modules/.pnpm/@smithy+eventstream-serde-node@2.1.3/node_modules/@smithy/eventstream-serde-node/dist-es/EventStreamMarshaller.js:2:25:
      2 │ import { Readable } from "stream";
        ╵                          ~~~~~~~~

  The package "stream" wasn't found on the file system but is built into node.
  Add "node_compat = true" to your wrangler.toml file and make sure to prefix the module name with "node:" to enable Node.js compatibility.
```

Code-Hex:
Even if you edit the contents of node_modules and modify these codes, you may encounter the following error.

```
✘ [ERROR] No matching export in "../../node_modules/.pnpm/@aws-sdk+credential-providers@3.525.0/node_modules/@aws-sdk/credential-providers/dist-es/index.browser.js" for import "fromNodeProviderChain"

    ../../node_modules/.pnpm/@anthropic-ai+bedrock-sdk@0.9.0/node_modules/@anthropic-ai/bedrock-sdk/auth.mjs:3:9:
      3 │ import { fromNodeProviderChain } from '@aws-sdk/credential-providers';
        ╵          ~~~~~~~~~~~~~~~~~~~~~
```

To resolve this, it was necessary to apply the following code:

```js
const signer = new SignatureV4({
        service: 'bedrock',
        region: props.regionName,
        credentials: {
                accessKeyId: props.awsAccessKey ?? "",
[...]

PR #474, opened 2024-07-23
feat(bedrock-sdk): allow for custom provider chain resolvers

linus-amg:
to allow also being executed in a non-node environment.

I'm trying to modify the bedrock sdk code so that it's executable in a non-node environment (e.g. vercel's edge runtime), as a first step I swapped the `fromNodeProviderChain` with a dynamic import and as a second step the client creation now allows for defining a custom provider chain resolver, which should enable the following usage:

```js
import AnthropicBedrock from '@anthropic-ai/bedrock-sdk';

const client = new AnthropicBedrock({
  awsSecretKey: '<insert your own aws secret key>',
  awsAccessKey: '<insert your own aws access key>',
  providerChainResolver: customFunctionWhichReturnsCredentialProvider
});
```

whenever a `providerChainResolver` is not defined, the `fromNodeProviderChain` from `@aws-sdk/credential-providers` should be used.

this should help fixing issue: https://github.com/anthropics/anthropic-sdk-typescript/issues/380 + https://github.com/anthropics/anthropic-sdk-typescript/issues/314#issuecomment-1980259476

I'm looking for a way to automatically test this new implementation, any ideas? I saw that for testing the bedrock sdk there is a script called "test" which executes: `bin/check-test-server && yarn jest` but `bin/check-test-server` does not exist in the public repository, are there other ways to test this?


RobertCraigie:
Thanks for the PR!

> I'm trying to modify the bedrock sdk code so that it's executable in a non-node environment (e.g. vercel's edge runtime), as a first step I swapped the fromNodeProviderChain with a dynamic import

~Is this strictly required? e.g. did you run into an error when this import wasn't dynamic?~

I'm asking as we want to support both CJS & ESM and I have a suspicion that dynamic import won't work with CJS, we'd need to double check.

edit: the comment you [linked](https://github.com/anthropics/anthropic-sdk-typescript/issues/314#issuecomment-1980259476) does show that a non-dynamic import causes breakage.

> I'm looking for a way to automatically test this new implementation, any ideas? I saw that for testing the bedrock sdk there is a script called "test" which executes: bin/check-test-server && yarn jest but bin/check-test-server does not exist in the public repository, are there other ways to test this?

Ah unfortunately we don't automated tests setup for the Bedrock SDK yet, we just rely on smoke tests running the example scripts.

linus-amg:
thank you for checking @RobertCraigie - so what's needed for this PR to move forward is to provide a way to make this work for cjs, correct?

RobertCraigie:
> so what's needed for this PR to move forward is to provide a way to make this work for cjs, correct?

yes, I think we won't be able to merge this without supporting CJS unfortunately

RobertCraigie:
I think we might be able to use export maps like we [do](https://github.com/anthropics/anthropic-sdk-typescript/blob/b7e21131de5cf9d741e3722393285b35eff26c1e/package.json#L66) in the main SDK package right? then users don't have to know to pass a custom provider chain resolver

gaodeng:
any news?

RobertCraigie:
Sorry for the delay here, I think I was incorrect in my original comment and this approach should work. We've pushed an updated version for the next release https://github.com/anthropics/anthropic-sdk-typescript/pull/790

Thanks Linus for the original PR!

PR #502, opened 2024-08-21
fix(docs): wrong method raw-streaming example

KernelDeimos:


RobertCraigie:
Thanks for the PR but `.create()` is the right method here as this example is specifically for the "raw" streaming case, i.e. not using the SDK-level helpers.

I'm curious where this confusion came from, maybe there are some other docs that need updating?

Issue #346, opened 2024-03-20
"Error: [object Object]" during message streaming when error is via an SSE (cause/detail not accessible)


paulcalcraft:
When hitting an error during the async iterator of a `anthropic.messages.create()`, the exception raised and associated error object doesn't have any detail, it just has an e.cause.message set to `"[object Object]"`.

My example SSE that's occurring during streaming is:

```javascript
{event: 'error', data: '{"type":"error","error":{"type":"overloaded_error","message":"Overloaded"}              }', raw: Array(2)}
```

The error SSE is then thrown using APIError.generate here: https://github.com/anthropics/anthropic-sdk-typescript/blob/ad92b0d536508954ee8b6c83e82bca30eefeb298/src/streaming.ts#L95

The errJSON is correctly being passed to generate, but because status isn't set (it's an SSE, not an HTTP response), we use castToError() to raise the APIConnectionError, with no other info:
https://github.com/anthropics/anthropic-sdk-typescript/blob/ad92b0d536508954ee8b6c83e82bca30eefeb298/src/error.ts#L52

castToError just returns new Error(errJSON)
https://github.com/anthropics/anthropic-sdk-typescript/blob/ad92b0d536508954ee8b6c83e82bca30eefeb298/src/core.ts#L977

But errJSON doesn't have a good toString, so our cause Error object has message "[object Object]" and no other properties. This means you can't handle/inspect the error cause correctly when catching the error during the async iterator.

An example error:
```
anthropic stream error Error: Connection error.
    at APIError.generate (\node_modules\@anthropic-ai\sdk\error.mjs:32:20)
    at Stream.iterator (\node_modules\@anthropic-ai\sdk\streaming.mjs:73:40)
    at process.processTicksAndRejections (\lib\internal\process\task_queues.js:95:5)
    at async [Symbol.asyncIterator] (\src\routes\api\oracle\+server.js:205:38)
...
    at async initiateResponse (/src/routes/api/oracle/+server.js:1410:17) {status: undefined, headers: undefined, error: undefined, cause: Error: [object Object]
    at castToError (fil…/node_modules/@anthr…, stack: 'Error: Connection error.
    at APIError.gene…/src/routes/api/oracle/+server.js:1410:17)', …}
```

And where I'm catching it:
```javascript
try {
    for await (const messageStreamEvent of response) {
        // ...
    }
} catch (e) {
    console.error("anthropic stream error", e) // e.cause.message == "[object Object]"
}
```

[...]

PR #503, opened 2024-08-21
chore: better object fallback behaviour for casting errors

greg84:
#346


RobertCraigie (approved):
Thanks for the PR! We fixed the error message serialisation separately in [this PR](https://github.com/anthropics/anthropic-sdk-typescript/pull/524) but I don't see a reason not to include this fallback behaviour.

PR #654, opened 2025-01-13
feat(stream): add `.withResponse()`

RobertCraigie:
make it more consistent with the rest of the API


michael-cohen-io (approved):
do we need this on the other SDKs as well?

RobertCraigie:
The Python SDK already supports accessing the raw response object with `.response` and the Go SDK only provides an accumulation helper so the default raw response API will just work :)

Issue #645, opened 2024-12-20
New builtin fetch alpha


RobertCraigie:
We've been working on an alpha version of the SDK that alongside many other cleanups, migrates from `node-fetch` to builtin fetch.

You can try it out by installing the `0.34.0-alpha.0` version:

```sh
pnpm add @anthropic-ai/sdk@alpha
```

The core happy path of the SDK remains the same, your existing method calls should just work! However you may run into typing issues as we now expect that `fetch` is defined as a global.

There are many other small breaking changes which you may run into, we'll share a full migration guide as soon as possible.

---

Please share feedback on anything you think could be improved or if you run into any issues! Hope this is a nice early present for you all :)

> [!IMPORTANT]  
> Unfortunately we cannot guarantee support over the holiday break (December 20th - January 6th inclusive) but we will go through every comment as soon as possible.

Nishchit14:
[ky](https://github.com/sindresorhus/ky) is a great replacement of fetch (it's a wrapper with great DX., sharing here incase you'd like to consider it.

RobertCraigie:
Thanks! However we're not considering other HTTP libraries as one great benefit of just depending on the builtin `fetch` function is that, among other things, we now have zero dependencies!

Nishchit14:
Agree, But ky has zero deps and works best for browser, Node, Bun etc. Just a suggestion.

![Image](https://github.com/user-attachments/assets/a24e48cb-105e-43f5-898d-8309134221e2)

pelikhan:
do you have a migration guide to support httpAgent from https://www.npmjs.com/package/https-proxy-agent ?

RobertCraigie:
@pelikhan we haven't tested `https-proxy-agent` directly yet but here's what setting a proxy would look like in a couple different runtimes

Node:
```ts
import Anthropic from '@anthropic-ai/sdk';
import * as undici from 'undici';

const dispatcher = new undici.ProxyAgent('http://localhost:8888');

const client = new Anthropic({
  fetch(url, init = {}) {
    return undici.fetch(url, {
      ...(init as unknown as undici.RequestInit),
      dispatcher,
    }) as unknown as Promise<Response>;
  },
});
```

Bun:
```ts
import Anthropic from '@anthropic-ai/sdk';

const client = new Anthropic({
  fetch(url, init) {
    return fetch(url, { ...init, proxy: 'http://localhost:8888' });
  },
});
```

Deno:
```ts
[...]

PR #715, opened 2025-03-11
Encode forward slashes in ARNs

clayallsopp:
The underlying issue in https://github.com/anthropics/claude-code/issues/297 is because inference profile ARNs contain a `/`, but these are not being encoded properly right now and the generated path is invalid.

I did a simple string replace, because I'm not sure the rest of the ARN needs to be encoded, but happy to change as necessary!

Example:

```
import { AnthropicBedrock } from '@anthropic-ai/bedrock-sdk';

const client = new AnthropicBedrock();

async function main() {
  const message = await client.messages.create({
    model: process.env.MODEL,
    messages: [
      {
        role: 'user',
        content: 'Hello!',
      },
    ],
    max_tokens: 1024,
  });
  console.log(message);
}

main();
```

Before:

```
$ DEBUG=true MODEL=arn:aws:bedrock:us-east-1:id:application-inference-profile/xxx node index.js

Anthropic:DEBUG:request https://bedrock-runtime.us-east-1.amazonaws.com/model/arn:aws:bedrock:us-east-1:id:application-inference-profile/xxx/invoke {
...
```

After:

```
$ DEBUG=true MODEL=arn:aws:bedrock:us-east-1:id:application-inference-profile/xxx node index.js

Anthropic:DEBUG:request https://bedrock-runtime.us-east-1.amazonaws.com/model/arn:aws:bedrock:us-east-1:id:application-inference-profile%2Fxxx/invoke {
...
```

### DIFF
--- a/packages/bedrock-sdk/src/client.ts
+++ b/packages/bedrock-sdk/src/client.ts
@@ -141,8 +141,12 @@ export class AnthropicBedrock extends Core.APIClient {
         throw new Error('Expected request body to be an object for post /v1/messages');
       }
 
-      const model = options.body['model'];
+      let model = options.body['model'];
       options.body['model'] = undefined;
+      // Encode any forward slashes in inference profile ARNs
+      if (model.includes('/')) {
+        model = model.replaceAll('/', '%2F');
+      }
 
       const stream = options.body['stream'];
       options.body['stream'] = undefined;


RobertCraigie:
Thanks for the PR! We're going to be releasing a new version of the SDK very soon that does this by default as it's a breaking change.

https://github.com/anthropics/anthropic-sdk-typescript/issues/645

This will break anyone who was manually encoding the URI themselves so I'm going to close this and this issue will be fixed properly in the upcoming release :)

PR #722, opened 2025-03-20
update alpha custom code for bedrock and vertex

stainless-em:


packages/bedrock-sdk/src/client.ts
```diff
@@ -1,13 +1,20 @@
...
 import { getAuthHeaders } from './auth';
 import { Stream } from './streaming';
+import { readEnv } from './internal/utils/env';
+import { FinalRequestOptions } from './internal/request-options';
+import { isObj } from './internal/utils/values';
+import { buildHeaders, type HeadersLike } from './internal/headers';
+import { FinalizedRequestInit } from './internal/types';
+import type { Fetch } from './internal/builtin-types';
+import type { MergedRequestInit } from './internal/types';
+export { BaseAnthropic } from '@anthropic-ai/sdk/client';
```
RobertCraigie, line 13: super nit: I find mixing and matching imports/exports like this difficult to read, should at least include a new line imo
```suggestion

export { BaseAnthropic } from '@anthropic-ai/sdk/client';
```

packages/bedrock-sdk/src/index.ts
```diff
@@ -1,4 +1 @@
-import { AnthropicBedrock } from './client';
-
-export default AnthropicBedrock;
-export { AnthropicBedrock };
+export { AnthropicBedrock, AnthropicBedrock as default } from './client';
```
RobertCraigie: aside: we should probably export a couple more things here, e.g. `ClientOptions`?

packages/vertex-sdk/examples/vertex.ts
```diff
@@ -14,7 +14,7 @@ async function main() {
         content: 'Hello!',
       },
     ],
-    model: 'claude-3-5-sonnet-v2@20241022',
+    model: 'claude-3-7-sonnet@20250219',
```
RobertCraigie: aside: might as well update more examples to use this model

packages/vertex-sdk/src/client.ts
```diff
@@ -1,13 +1,19 @@
...
+import { readEnv } from './internal/utils/env';
+import { FinalRequestOptions } from './internal/request-options';
+import { FinalizedRequestInit } from './internal/types';
+import { isObj } from './internal/utils/values';
+import { buildHeaders, HeadersLike } from './internal/headers';
+export { BaseAnthropic } from '@anthropic-ai/sdk/client';
+import type { Fetch } from './internal/builtin-types';
+import type { MergedRequestInit } from './internal/types';
 
 const DEFAULT_VERSION = 'vertex-2023-10-16';
```
RobertCraigie, line 14: intentional that `BaseAnthropic` isn't exported here but it is for bedrock?

src/internal/headers.ts
```diff
@@ -9,7 +9,9 @@ export type HeadersLike =
   | null
   | NullableHeaders;
 
-const brand_privateNullableHeaders = Symbol('brand.privateNullableHeaders');
+const brand_privateNullableHeaders = Symbol.for('brand.privateNullableHeaders') as symbol & {
+  description: 'brand.privateNullableHeaders';
+};
```
RobertCraigie, line 14: nit: we should update codegen to output this as well right?

RobertCraigie (approved):
Looks great!

Will need to make sure that the `postprocess-dist-package-json.cjs` scripts are updated to include the new version in the minimum range, doesn't necessarily have to be in this PR though

packages/vertex-sdk/src/client.ts
stainless-em: it is 😭
i will rearrange lol

src/internal/headers.ts
stainless-em: this is pretty hacky tbh and i only did it here because there wasn't really an alternative...
it's only relevant for subclassing from external packages anyway which ~nobody else should be doing

RobertCraigie: yeah that is fair, I was just thinking from the custom code maintenance perspective - ideally we'd have as small as possible of a divergence with custom code editing our internals to reduce the merge conflicts burden.

but this should rarely be edited and I don't feel strongly

packages/vertex-sdk/examples/vertex.ts
stainless-em: maybe i should revert instead and let them decide when they want to change them all?

RobertCraigie: yeah sure

PR #772, opened 2025-05-23
fix: correct streaming error message to link to TypeScript SDK docs

jerhadf:
Fixed incorrect documentation link in streaming timeout error message that was pointing to Python SDK docs instead of TypeScript SDK docs


RobertCraigie:
thanks for the PR! sorry we missed it until now, this has since been fixed.

PR #795, opened 2025-07-23
example:  add example for image input understanding

princeaden1:
### Summary

This PR introduces an example for using image inputs with the Anthropic client. It enables multimodal functionality where the model can understand and respond to image-based prompts alongside text.

### Changes

- Extended client to accept image files as input (e.g., via `fs.createReadStream`)
- Added example usage in `examples/images.ts` demonstrating how to send an image prompt

### Usage

```ts
const filePath = path.join(__dirname, 'logo.png');
  const buffer = fs.readFileSync(filePath);
  const base64Image = buffer.toString('base64');
  const result = await client.messages.create({
    messages: [
      {
        role: 'user',
        content: [
          {
            type: 'text',
            text: 'What is in this image?',
          },
          {
            type: 'image',
            source: {
              data: base64Image,
              type: 'base64',
              media_type: 'image/png',
            },
          },
        ],
      },
    ],
    model: 'claude-sonnet-4-20250514',
    max_tokens: 1024,
  }); ```


localden:

A few issues prevent this from landing as-is:

- `logo.png` is committed to the repo root but the example reads `path.join(__dirname, 'logo.png')`, which resolves to `examples/logo.png` and would fail with ENOENT
- The model ID `claude-sonnet-4-20250514` is outdated — current examples use `claude-sonnet-4-6` (e.g. [`examples/agents.ts:17`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/examples/agents.ts#L17))
- The PR description says "Extended client to accept image files as input," but no SDK code is changed — this only uses the existing [`ImageBlockParam`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/resources/messages/messages.ts) types

Closing for now. If you'd like to resubmit, a self-contained example using a public image URL (rather than a committed binary) with a current model ID would be welcome.

PR #801, opened 2025-08-04
update vertex base URL

### LINKED ISSUE #800: Vertex base URL is incorrect when using the global endpoint
Per [Google's docs](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations), the hostname when using the global endpoint on Vertex should be `aiplatform.googleapis.com`.

The current code in `packages/vertex-sdk/src/client.ts` has:

```
    super({
      baseURL: baseURL || `https://${region}-aiplatform.googleapis.com/v1`,
      ...opts,
    });
```

This results in the hostname being `global-aiplatform.googleapis.com`, which does not work (404 error).

The workaround is for the caller to set baseURL themselves, but obviously the SDK should handle this automatically as it is very confusing to get a 404 this way.

choir241:
update vertex baseURL to match Google docs so global endpoint is `aiplatform.googleapis.com` instead of the current URL `global-aiplatform.googleapis.com`, which returns a 404 error
 
Closes #800


RobertCraigie:
Thanks for the PR! We actually had a fix queued up for this already it just hadn't been released when you opened this PR. It has since been released.

PR #812, opened 2025-08-28
docs: fix default timeout comment

aaronmaxlevy:
Correcting comment related to default timeout calculation. Currently `calculateNonstreamingTimeout` has `const maxTime = 60 * 60 * 1000; // 10 minutes` which is misleading since that is actually 60 minutes :)


RobertCraigie (approved):
🤦 thanks!

aaronmaxlevy:
No problem! :)

PR #826, opened 2025-10-06
fix(vertex): extract model from request body for count_tokens endpoint

### LINKED ISSUE #593: countTokens not working with @anthropic-ai/vertex-sdk
Hi, I encountered this issue today. I tried updating to the latest version, but it's still not working.

Here is some information about the library:

Name: @anthropic-ai/vertex-sdk
Version: 0.5.2
I am trying to count the tokens of a message when using AnthropicVertex.

image

However, it returns an error:

image

I checked in the library, and it seems to be calling the endpoint v1/messages/count_tokens?beta=true.

image

Does anyone have any ideas on how to resolve this issue? Thank you for reading this

ChristianRyan:
## Problem
The `countTokens()` method on Vertex AI fails with "Requested entity was not found" error.

**Root cause:** The SDK hardcodes `"count-tokens"` as the model name in the API path instead of using the actual model from the request body.

## Reproduction
```typescript
const client = new AnthropicVertex({ region: 'us-east5' });
await client.beta.messages.countTokens({
  model: 'claude-sonnet-4-5@20250929',  // User provides actual model
  messages: [{ role: 'user', content: 'test' }],
});
// Error: Requested entity was not found
// Because SDK calls: .../models/count-tokens:rawPredict
// Instead of: .../models/claude-sonnet-4-5@20250929:rawPredict
```

## Solution
Extract the model from `options.body['model']` and use it in the path construction, matching the pattern used by the regular messages endpoint (lines 151-170).

## Changes
- Extract model from request body in count_tokens handler
- Add validation for request body object
- Use actual model in path construction instead of hardcoded "count-tokens"

## Testing
- [x] Existing tests pass
- [ ] Manual testing with real Vertex AI project (requires Vertex credentials)

## Related
- Fixes #593

### DIFF
--- a/packages/vertex-sdk/src/client.ts
+++ b/packages/vertex-sdk/src/client.ts
@@ -179,7 +179,13 @@ export class AnthropicVertex extends BaseAnthropic {
         );
       }
 
-      options.path = `/projects/${this.projectId}/locations/${this.region}/publishers/anthropic/models/count-tokens:rawPredict`;
+      if (!isObj(options.body)) {
+        throw new Error('Expected request body to be an object for post /v1/messages/count_tokens');
+      }
+
+      const model = options.body['model'];
+
+      options.path = `/projects/${this.projectId}/locations/${this.region}/publishers/anthropic/models/${model}:rawPredict`;
     }
 
     return super.buildRequest(options);


ChristianRyan:
Closing this PR - further investigation shows the real issue is that count_tokens is only supported in specific regions (us-east5, europe-west1, asia-east1, asia-southeast1, us-central1, europe-west4) and NOT in the 'global' region. The error occurs when users have region='global' configured. A different fix is needed.

PR #849, opened 2025-11-18
fix(client): don't strip path from filename for skills endpoint

sd-st:


packyg (commented):
I think this is undesired when using `toFile` for the files API as that endpoint expects the filename to not be a path

I guess you could say it's caller's responsibility to name the file correctly, but I'm somewhat concerned about breaking existing use cases

sd-st:
Yeah that is totally reasonable. It is a little tricky to do this in a way that's intuitive but also doesn't break existing code. Like we could omit stripping the file name for the skills endpoint but leave it in for other endpoints, though that may be confusing to some 🤷‍♀️

packyg:
> Yeah that is totally reasonable. It is a little tricky to do this in a way that's intuitive but also doesn't break existing code. Like we could omit stripping the file name for the skills endpoint but leave it in for other endpoints, though that may be confusing to some 🤷‍♀️

@sd-st I think it could be confusing yeah but it reflects API usage. Could we make it so that we apply this change (don't strip paths) and update the file method itself to do the path stripping to preserve existing behavior?

sd-st:
> > Yeah that is totally reasonable. It is a little tricky to do this in a way that's intuitive but also doesn't break existing code. Like we could omit stripping the file name for the skills endpoint but leave it in for other endpoints, though that may be confusing to some 🤷‍♀️
> 
> @sd-st I think it could be confusing yeah but it reflects API usage. Could we make it so that we apply this change (don't strip paths) and update the file method itself to do the path stripping to preserve existing behavior?

Hey, sorry for missing this! That totally makes sense to me. I'll update this PR shortly

src/internal/uploads.ts
```diff
@@ -49,20 +49,17 @@ export function makeFile(
...
-        (('name' in value && value.name && String(value.name)) ||
-          ('url' in value && value.url && String(value.url)) ||
-          ('filename' in value && value.filename && String(value.filename)) ||
-          ('path' in value && value.path && String(value.path)))) ||
-      ''
-    )
-      .split(/[\\/]/)
-      .pop() || undefined
-  );
+export function getName(value: any, strip: boolean): string | undefined {
```
RobertCraigie: nit: current name is not super clear tbh
```suggestion
export function getName(value: any, stripPath: boolean): string | undefined {
```

src/internal/to-file.ts
```diff
@@ -90,7 +90,7 @@ export async function toFile(
   // If it's a promise, resolve it.
   value = await value;
 
-  name ||= getName(value);
+  name ||= getName(value, true);
```
packyg, line 93: Do the other packages (bedrock/foundry/vertex) need this change too? They also have a `toFile` helper

packyg (commented):
One question and agree with @RobertCraigie 's comment


But I like the updated interface where the caller doesn't have to think about whether to strip

src/internal/to-file.ts
sd-st: I don't think so - looks like they just symlink back to this file

packyg (approved):

PR #853, opened 2025-11-21
refactor(structured outputs): rename `parsed` to `parsed_output`

RobertCraigie:
This was an accident, we should use `.parsed_output` to make it more clear what it does, and to match the other SDKs.


RobertCraigie:
cc @karpetrosyan

karpetrosyan:
lgtm, minor question: when should we use camel case and when snake case in typescript? I used to think we should always go with camel case here

karpetrosyan:
One more minor note: maybe it would be better to commit this as 'refactor' instead of 'fix', 
something like:

```
refactor(structured-outputs): rename `parsed` to `parsed_output` for text blocks
```

RobertCraigie:
We use snake case for things like this so it's more consistent, as the API uses snake case

Issue #829, opened 2025-10-12
`messages.stream().withResponse()` throws uncatchable errors


soorria:
When using `.withResponse()` with the Anthropic streaming API, errors are caught by surrounding try-catch blocks BUT are also lead to uncaught exceptions that crash the Node.js process.

Happy to take a crack at fixing it.

## Reproduction Steps

See github repo [here](https://github.com/soorria/anthropic-sdk-bug), but tldr:

this issue happens with any stream calls that look like the following, as long as the response errors.

```js
const { data: stream } = await anthropic.messages
  .stream({
    max_tokens: 1,
    messages: [],
    model: 'claude-sonnet-4-5',
  })
  .withResponse()
```

This does not happen when using `.create` or when streaming without `.withResponse`

### Expected Behavior

Errors should be caught by the try-catch block, preventing uncaught exceptions.

### Actual Behavior

The same error is both:

1. Caught by the try-catch block
2. **Also** thrown as an uncaught exception, killing the node process or triggering `process.on('uncaughtException')` handlers.

## Misc Details

Anthropic SDK version: `0.65.0`
Node version: `v22.13.1`

soorria:
Looks like this is _kinda_ intended behaviour according to [this comment](https://github.com/anthropics/anthropic-sdk-typescript/blob/d7bab45cd32a40fe32923743267878d3e1f8d520/src/lib/MessageStream.ts#L402-L410). This doesn't seem to handle when `.withResponse()` is called tho, so there's a fairly simple solution.

<img width="838" height="212" alt="Image" src="https://github.com/user-attachments/assets/79c1969a-1adb-49f9-acf8-d9e46da63fe6" />

soorria:
PRed a fix, but could also be good to wrap the uncaught error or at least update its stack trace. Right now the stack trace looks like:

```
AuthenticationError: 401 {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011CU3J65PFgFoMgJhoHSLJb"}
    at APIError.generate (file:///node_modules/.pnpm/@anthropic-ai+sdk@0.65.0/node_modules/@anthropic-ai/sdk/core/error.mjs:40:20)
    at Anthropic.makeStatusError (file:///node_modules/@anthropic-ai/sdk/client.mjs:152:32)
    at Anthropic.makeRequest (file:///node_modules/@anthropic-ai/sdk/client.mjs:306:30)
[...]

PR #856, opened 2025-12-03
fix(streams): ensure errors are catchable 

404Wolf:
Fix https://github.com/anthropics/anthropic-sdk-typescript/issues/829

### DIFF
--- a/src/lib/MessageStream.ts
+++ b/src/lib/MessageStream.ts
@@ -112,6 +112,8 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
     response: Response;
     request_id: string | null | undefined;
   }> {
+    this.#catchingPromiseCreated = true;
+
     const response = await this.#connectedPromise;
     if (!response) {
       throw new Error('Could not resolve a `Response` object');
@@ -133,7 +135,7 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
    */
   static fromReadableStream(stream: ReadableStream): MessageStream {
     const runner = new MessageStream();
-    runner._run(() => runner._fromReadableStream(stream));
+    void runner._run(() => runner._fromReadableStream(stream));
     return runner;
   }
 
@@ -146,21 +148,24 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
     for (const message of params.messages) {
       runner._addMessageParam(message);
     }
-    runner._run(() =>
-      runner._createMessage(
-        messages,
-        { ...params, stream: true },
-        { ...options, headers: { ...options?.headers, 'X-Stainless-Helper-Method': 'stream' } },
-      ),
-    );
+    void runner
+      ._run(
+        async () =>
+          await runner._createMessage(
+            messages,
+            { ...params, stream: true },
+            { ...options, headers: { ...options?.headers, 'X-Stainless-Helper-Method': 'stream' } },
+          ),
+      )
+      .catch(runner.#handleError);
     return runner;
   }
 
-  protected _run(executor: () => Promise<any>) {
-    executor().then(() => {
-      this._emitFinal();
-      this._emit('end');
-    }, this.#handleError);
+  protected async _run(executor: () => Promise<any>) {
+    await executor();
+
+    this._emitFinal();
+    this._emit('end');
   }
 
   protected _addMessageParam(message: MessageParam) {
--- a/src/lib/MessageStream.ts
+++ b/src/lib/MessageStream.ts
@@ -135,7 +135,7 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
    */
   static fromReadableStream(stream: ReadableStream): MessageStream {
     const runner = new MessageStream();
-    void runner._run(() => runner._fromReadableStream(stream));
+    runner._run(() => runner._fromReadableStream(stream));
     return runner;
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/MessageStream.ts
```diff
@@ -146,21 +148,24 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
...
+            { ...params, stream: true },
+            { ...options, headers: { ...options?.headers, 'X-Stainless-Helper-Method': 'stream' } },
+          ),
+      )
+      .catch(runner.#handleError);
     return runner;
   }
 
-  protected _run(executor: () => Promise<any>) {
-    executor().then(() => {
```
404Wolf, line 160: Previously this would "leak" exceptions. It did async stuff but wasn't async, so we couldn't .catch anything that happened inside

examples/debugging.ts
RobertCraigie, line 1: accidentally left in?

404Wolf: Whoops yes! I just removed

src/lib/MessageStream.ts
```diff
@@ -146,21 +148,24 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
...
-      runner._createMessage(
-        messages,
-        { ...params, stream: true },
-        { ...options, headers: { ...options?.headers, 'X-Stainless-Helper-Method': 'stream' } },
-      ),
-    );
+    void runner
+      ._run(
+        async () =>
+          await runner._createMessage(
```
controversial: async function with a single `return await` in the body is equivalent to the previous behavior; you should be able to just return the promise directly

```suggestion
    runner
      ._run(() =>
        runner._createMessage(
```

src/lib/MessageStream.ts
```diff
@@ -146,21 +148,24 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
...
       runner._addMessageParam(message);
     }
-    runner._run(() =>
-      runner._createMessage(
-        messages,
-        { ...params, stream: true },
-        { ...options, headers: { ...options?.headers, 'X-Stainless-Helper-Method': 'stream' } },
-      ),
-    );
+    void runner
```
controversial: `void` has no runtime effect; it’s typically just used as an `eslint-disable` comment for `@typescript-eslint/no-floating-promises` 

in this case you don’t have a floating promise because you have a `.catch()`, so I think you should be able to remove the `void` without getting a lint error

src/lib/MessageStream.ts
```diff
@@ -133,7 +135,7 @@ export class MessageStream implements AsyncIterable<MessageStreamEvent> {
    */
   static fromReadableStream(stream: ReadableStream): MessageStream {
     const runner = new MessageStream();
-    runner._run(() => runner._fromReadableStream(stream));
+    void runner._run(() => runner._fromReadableStream(stream));
```
controversial: i don’t think `no-floating-promises` is enabled on this project actually so you can remove this `void` as well. maybe add a .catch unless you’re sure that `_fromReadableStream` won’t error

src/lib/MessageStream.ts
controversial: didn’t `this.#handleError`—passed for `onRejected`—handle anything that happened inside in the old version? I’m not sure I understand how the new code is functionally different from the old

it seems like this PR is just lifting the unawaited async code up a level but I’m not sure I can see where the functional difference is for exception handling

src/lib/MessageStream.ts
controversial: (previously, I think this invocation _did_ have an error handler in the `#handleError` `onRejected` within `runner.run`?)

tests/api-resources/MessageStream.test.ts
```diff
@@ -225,4 +225,31 @@ describe('MessageStream class', () => {
     assertToolUseResponse(events, finalMessage);
     expect(finalText).toBe("I'll check the current weather in Paris for you.");
   });
+
+  it('does not throw unhandled rejection with withResponse()', async () => {
```
controversial, line 229: Is this a test case that fails on `main`? for me this test passes with the old version of `MessageStream.ts`
```sh
> git checkout main src/lib/MessageStream.ts
> jest -t "does not throw unhandled rejection with withResponse"
Tests:       420 skipped, 1 passed, 421 total
```
unless i’m missing something about how to set up my environment, it would be good to have a test that reliably fails to demonstrate the previously-unfixed behavior

404Wolf: It does fail on main, the issue is the test was testing beta message stream. I'll change it to test the normal message stream. The fix was the same for both anyway

404Wolf:
@controversial I was sort of fooling myself, the main thing here is just setting `this.#catchingPromiseCreated = true;` so that we know in `#handleError`. I like being explicit about async/await but yes it's besides the point. How does this look?

tests/api-resources/MessageStream.test.ts
404Wolf: If you wanted to see it fail on main you'd need to have done `git checkout main src/lib/BetaMessageStream.ts`. But regardless this test should be for the non beta one

controversial (approved):
This looks great, thanks for pointing me in the right direction with the test case! sorry for the back & forth on the promise handling. nice work !

PR #859, opened 2025-12-09
feat(errors): add ignore_tool_parameter_parse_errors

manucorporat:
…ageStream to handle parse errors

- Introduced UnableToParseToolParameterError for better error handling.
- Updated BetaMessageStream to conditionally ignore tool parameter parse errors based on new ignore_tool_parameter_parse_errors flag in MessageCreateParamsBase.
- Enhanced error handling during JSON parsing in BetaMessageStream.

### DIFF
--- a/src/core/error.ts
+++ b/src/core/error.ts
@@ -4,6 +4,8 @@ import { castToError } from '../internal/errors';
 
 export class AnthropicError extends Error {}
 
+export class UnableToParseToolParameterError extends AnthropicError {}
+
 export class APIError<
   TStatus extends number | undefined = number | undefined,
   THeaders extends Headers | undefined = Headers | undefined,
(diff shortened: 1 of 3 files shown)


sanyamkamat (approved):

pabloelisseo (approved):

RobertCraigie (commented):
I don't think we should ship a `ignore_tool_parameter_parse_errors` parameter.

Instead it would be a much better API if the `error` callback handler could somehow tell the stream to keep going.

e.g. something like this
```typescript
stream.on('error', (error) => {
  if (error instanceof UnableToParseToolParameterError) {
    return { continue: true };
  }
});
```

---

The separate `UnableToParseToolParameterError` error class is a good change 👍.

manucorporat:
i can make this change!

dtmeadows-ant:
@manucorporat thanks. Per Robert's review, we don't want a request parameter for what is client-side parse leniency. `messages.ts` is generated from the API spec, and a param there has to be something the API itself understands. If tolerating malformed tool JSON mid-stream is still something you need, the direction we'd consider is letting the stream's error handler opt to continue. An issue sketching that would be the place to start.

Issue #292, opened 2024-02-14
"Unexpected end of JSON input" when streaming on edge environments (Vercel Edge, Cloudflare Workers)


venables:
The SDK seems to operate fine when running in a node.js environment, but when running in an Edge runtime (browser env), such as Vercel Edge or Cloudflare Workers, streaming becomes cut off with the following exception:

```
Could not parse message into JSON: 
From chunk: [ 'event: content_block_delta' ]

SyntaxError: Unexpected end of JSON input
    at (node_modules/@anthropic-ai/sdk/streaming.mjs:58:39)
    at (app/api/test/route.js:15:19)
    at (node_modules/next/dist/esm/server/future/route-modules/app-route/module.js:189:36)
    at (node_modules/next/dist/esm/server/future/route-modules/app-route/module.js:128:25)
    at (node_modules/next/dist/esm/server/future/route-modules/app-route/module.js:251:29)
    at (node_modules/next/dist/esm/server/web/edge-route-module-wrapper.js:81:20)
    at (node_modules/next/dist/esm/server/web/adapter.js:157:15)
```

The error is coming from this block: https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/streaming.ts#L69-L84

The line content is:

```js
{
  event: 'content_block_delta',
  data: '',
  raw: [ 'event: content_block_delta' ]
}
```

Since the data is an empty string,, the JSON parsing blows up.  I can bypass this error if I modify the code to ignore empty strings, but that does not seem ideal.  

## Reproduction repos:

I put the Streaming example from the [Anthropic SDK README](https://github.com/anthropics/anthropic-sdk-typescript?tab=readme-ov-file#streaming-responses) into a Vercel Edge function and a Cloudflare Workers function with the same failing result.

Note, the error occurs whether we use `import "@anthropic-ai/sdk/shims/web";` or not.

### Vercel Edge:

I've put together a sample repo, using create-next-app and using the example from your README: https://github.com/venables/anthropic-edge-stream-error 

The file in question would be [`app/api/test/route.ts`](https://github.com/venables/anthropic-edge-stream-error/blob/main/app/api/test/route.js).  If you remove `export const runtime = "edge"`, it works as expected.

This error will not occur locally since locally the environment is a node.js environment, but when you deploy to Vercel (with `runtime = "edge"` still in the code), you will consistently get the error.  

### Cloudflare Workers

[...]

PR #861, opened 2025-12-15
skip events with no data while streaming

fvdemand:
In my cloud environment while using v0.71.2

I have continued to have the issue described [here](https://github.com/anthropics/anthropic-sdk-typescript/issues/292).

I have seen it occur both with content_block_delta and content_block_stop. 

Somewhat of note I have not got this running locally.  

Logs: 
```
Could not parse message into JSON:
From chunk: [ 'event: content_block_delta' ]
```
I believe this is somewhat normal server behavior when the server is responding for a long time basically the raw string for the sse shows that there is no data. I added a condition to skip this event only when streaming.

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -66,6 +66,11 @@ export class Stream<Item> implements AsyncIterable<Item> {
             sse.event === 'content_block_delta' ||
             sse.event === 'content_block_stop'
           ) {
+            // skip streaming events with no data
+            if(sse.data == null) {
+              continue;
+            }
+
             try {
               yield JSON.parse(sse.data);
             } catch (e) {


felixfbecker:
@fvdemand sorry to hear you're having issues. It seems like what you're describing is either a bug in our API (in which case we need to fix it there), or a bug in the SSE parsing. I wouldn't expect `null` data to reach this part of the code, so I'm hesitant to add a workaround here without solving the underlying cause. Do you have an example script that reproduces this issue so we can investigate?

fvdemand:
I don't have a script that will be useful for you to reproduce. Here is the last bit that gave the error. However, I had the same issue when using the .stream and .finalMessage functions. 

```
const getFullStreamedResponseClaude = async (completionContext) => {
  completionContext = {
    ...completionContext,
    stream: true
  };
  const stream = await anthropicClient.messages.create(completionContext);
  try {
    const chunks = [];
    for await (const chunk of stream) {
      if (chunk.type === 'content_block_delta') {
        const text = chunk.delta?.text || '';
        chunks.push(text);
      }
    }
    return chunks.join('');
  } catch(error) {
    if(error instanceof TypeError) {
      console.log('Type error for stream: ', stream);
    }
    throw error;
  }
};
```

Do you agree that the log in my original message indicates that the Message code received a sse in which the .raw demonstrates that there is no .data ? For clarity those are not my logs, they come from in the .messages interface.

I will try to get a request ID. Update, I do not have access to the env is there a modification I can make to this block to get the request id?

felixfbecker:
I think we might have fixed this in the API or SDK recently, are you still seeing this issue in the latest version?

fvdemand:
I have been running with this change using an npm patch. I intend to get back on the main package. This week, I will run the scenario we originally saw this in without the patch and let you know if the issue persists

fvdemand:
@DukeDeSouth I read your message about why this is not the right fix. It seems to make sense to me

Issue #867, opened 2025-12-19
bug/proposal: infinitely hanging clients breaking bigger/complex sessions. Proposal: add Streaming Idle Timeout to Prevent Indefinite Hangs


notactuallytreyanastasio:
Hi all.

I have repeatedly been encountering a pretty frustrating bug on complex, longer sessions with Claude.

I did some digging with Claude's help and got to this issue report consulting the contribution guidelines etc and running through things as deeply as I could.

Please let me know if there is anything I need to change or process to follow that I may be missing.

I have some code to reference here as well.

This also seems to relate to issues #842 and #844 but goes a bit further with replication/explanation, and tries to offer some potential solutions and maybe illuminates a bit more with the real world example provided.

## Summary

When using streaming responses (`messages.stream()` or `messages.create({ stream: true })`), if the server stops sending SSE events mid-stream (due to network issues, server stalls, or connection problems), the client will wait indefinitely. There is currently no mechanism to detect and abort stalled streams.

### Current Behavior: A Real-World Failure

The SDK provides two timeout mechanisms:
- **`timeout`** option: Overall request timeout (default: 10 minutes)
- **`AbortController`**: Manual cancellation via signals

**Neither mechanism detects or handles stalled streams.** Here's what happened in a real production session:

#### The Incident

A Claude Code CLI session (`f35c7c15-802e-462b-b468-b000a96e40bb`) was performing a multi-step task using `claude-opus-4-5-20251101`. The session had been running successfully, accumulating 12,896 input tokens and 4,059 output tokens across multiple tool calls.

At 20:22:46 UTC, the API began streaming a response. The thinking block completed successfully:

```json
{
  "message": {
    "model": "claude-opus-4-5-20251101",
    "id": "msg_018BDP6sdKEzgJ5GS1rn7NW1",
    "content": [
      {
        "type": "thinking",
        "thinking": "I can see my active trace session `3c62b735`. It shows 12896 input tokens, 4059 output tokens..."
      }
    ],
    "stop_reason": null,    // <-- Stream still in progress
    "usage": {
      "output_tokens": 4,   // <-- Only 4 tokens output, thinking block only
      "cache_read_input_tokens": 73111
    }
  },
  "timestamp": "2025-12-19T20:22:46.441Z"
}
```

The response was actively streaming (`stop_reason: null`), but after the thinking block, **no more SSE events arrived**. The stream didn't close, didn't error, didn't timeout—it simply stopped sending data.

[...]

PR #868, opened 2025-12-19
bug/proposal: infinitely hanging clients breaking bigger/complex sessions. Proposal: add Streaming Idle Timeout to Prevent Indefinite Hangs #867

notactuallytreyanastasio:
# This is an example PR for what's brought up in #867 

I tried to ensure good test coverage.

## Summary

This PR adds an `idleTimeout` option to detect and abort streams that stop sending SSE events without closing the connection. This addresses a critical failure mode where streaming responses hang indefinitely when:

- Network connectivity becomes unstable mid-stream
- The API server experiences transient issues
- Load balancers or proxies close connections without proper cleanup

### The Problem

Currently, the SDK provides:
- `timeout`: Overall request timeout (covers connection + response time)
- `AbortController`: Manual cancellation via signals

Neither detects **stalled streams** where the connection is alive but no data is flowing. When this happens, `for await (const event of stream)` blocks forever with no error and no indication anything is wrong.

### Real-World Evidence

From production Claude Code CLI sessions, we observed:
- Sessions hanging for 15+ minutes with 0% CPU (waiting for data)
- `"stop_reason": null` in logs - response never completed
- Only solution was to kill the process, losing all work in progress

### The Solution

New `idleTimeout` option that:
1. Tracks time since the last SSE event was received
2. Uses `Promise.race` to race each chunk read against a timeout
3. Throws `StreamIdleTimeoutError` with diagnostic information if the stream stalls

## Changes

| File | Change |
|------|--------|
| `src/client.ts` | Add `idleTimeout` to `ClientOptions` |
| `src/internal/request-options.ts` | Add `idleTimeout` to `RequestOptions` |
| `src/core/error.ts` | Add `StreamIdleTimeoutError` class |
| `src/core/streaming.ts` | Implement timeout in `_iterSSEMessages` |
| `src/internal/parse.ts` | Pass options through to `Stream.fromSSEResponse` |
| `src/index.ts` | Export `StreamIdleTimeoutError` |
| `tests/streaming.test.ts` | Add 13 idle timeout unit tests |
| `tests/api-resources/MessageStream.test.ts` | Add 4 client-level integration tests |

## API

```typescript
// Set default idle timeout for all streams (90 seconds)
const client = new Anthropic({
  idleTimeout: 90_000,
});

// Override per-request
const stream = await client.messages.stream(params, {
  idleTimeout: 120_000, // 2 minutes for this specific request
});

// Catch the specific error type
try {
  for await (const event of stream) {
    // process events
  }
} catch (err) {
  if (err instanceof Anthropic.StreamIdleTimeoutError) {
    console.log(`Stream stalled after ${err.eventCount} events`);
    console.log(`Last event at: ${err.lastEventTime}`);
    console.log(`Configured timeout: ${err.idleTimeoutMs}ms`);
    // Retry or fail gracefully
  }
}
```

## Type Safety

- All new properties are optional (backward compatible)
- No `any` types used
- `StreamIdleTimeoutError` extends `APIConnectionError` (existing error hierarchy)
- `idleTimeout?: number | undefined` to satisfy `exactOptionalPropertyTypes`

## Test Plan

- [x] `npm run lint` passes
- [x] `npm test` passes (427 tests total, including 17 new idle timeout tests)
- [x] Types compile correctly

#### Low-Level Unit Tests (`streaming.test.ts`) - 13 tests

| Test | Purpose |
|------|---------|
| Basic timeout fires | Core functionality works |
| Timeout resets on data | Slow but consistent streams complete |
| Error has diagnostic info | `idleTimeoutMs`, `eventCount`, `lastEventTime` |
| No timeout without option | Backward compatibility |
| Options pass through Stream | Factory pattern works |
| Cleanup on normal completion | No timer memory leaks |
| Cleanup on user break | No leaks when breaking from loop |
| Cleanup on manual abort | No leaks on `controller.abort()` |
| Very short timeout (1ms) | Edge case behavior |
| Multiple sequential streams | No cross-contamination |
| Accurate event count | Diagnostic accuracy |
| No timer leak on processing error | Exception safety |

#### Client Integration Tests (`MessageStream.test.ts`) - 4 tests

| Test | Purpose |
|------|---------|
| Timeout via request options | `client.messages.stream(params, { idleTimeout })` |
| Timeout via client default | `new Anthropic({ idleTimeout })` |
| Request overrides client | Option precedence works |
| Normal completion with timeout | No false positives |

## Questions for Maintainers

1. **Default value**: Should `idleTimeout` have a default (e.g., 2 minutes) or remain opt-in?
2. **Ping events**: Should `ping` SSE events reset the idle timer, or only "meaningful" events?
3. **Retry integration**: Should idle timeout trigger retry logic if `maxRetries > 0`?

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -199,6 +199,22 @@ export interface ClientOptions {
    * @unit milliseconds
    */
   timeout?: number | undefined;
+
+  /**
+   * The maximum amount of time (in milliseconds) to wait between SSE events during streaming.
+   * If no event is received within this period, the stream will be aborted with a
+   * `StreamIdleTimeoutError`.
+   *
+   * This is distinct from `timeout` which applies to the overall request. `idleTimeout`
+   * specifically detects stalled streams where the connection is alive but no data is flowing.
+   *
+   * Can be overridden per-request via `RequestOptions.idleTimeout`.
+   *
+   * @unit milliseconds
+   * @default undefined (no idle timeout)
+   */
+  idleTimeout?: number | undefined;
+
   /**
    * Additional `RequestInit` options to be passed to `fetch` calls.
    * Properties will be overridden by per-request `fetchOptions`.
@@ -270,6 +286,7 @@ export class BaseAnthropic {
   baseURL: string;
   maxRetries: number;
   timeout: number;
+  idleTimeout: number | undefined;
   logger: Logger;
   logLevel: LogLevel | undefined;
   fetchOptions: MergedRequestInit | undefined;
@@ -314,6 +331,7 @@ export class BaseAnthropic {
 
     this.baseURL = options.baseURL!;
     this.timeout = options.timeout ?? BaseAnthropic.DEFAULT_TIMEOUT /* 10 minutes */;
+    this.idleTimeout = options.idleTimeout;
     this.logger = options.logger ?? console;
     const defaultLogLevel = 'warn';
     // Set default logLevel early so that we can log a warning in parseLogLevel.
@@ -342,6 +360,7 @@ export class BaseAnthropic {
       baseURL: this.baseURL,
       maxRetries: this.maxRetries,
       timeout: this.timeout,
+      idleTimeout: this.idleTimeout,
       logger: this.logger,
       logLevel: this.logLevel,
       fetch: this.fetch,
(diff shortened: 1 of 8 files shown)


localden:
Thank you for your contribution and the detailed write-up in issue [#867](https://github.com/anthropics/anthropic-sdk-typescript/issues/867)!

We're consolidating on [#959](https://github.com/anthropics/anthropic-sdk-typescript/pull/959), which implements the same `idleTimeout` feature with a more focused diff against [`src/core/streaming.ts`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/core/streaming.ts) and [`src/internal/request-options.ts`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/internal/request-options.ts).

Closing as duplicate of [#959](https://github.com/anthropics/anthropic-sdk-typescript/pull/959). Please continue the discussion there.

PR #874, opened 2026-01-04
fix(docs): remove broken @std/log link from README

jeremylongshore:
## Summary

The `jsr.io/@std/log` URL in the README returns 404. This PR removes the broken link while preserving all other working logger examples (pino, winston, bunyan, consola, signale).

## Test plan

- [x] Verified `jsr.io/@std/log` returns 404
- [x] Verified all other logger links return 200

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,7 +504,7 @@ may still be visible.
 #### Custom logger
 
 By default, this library logs to `globalThis.console`. You can also provide a custom logger.
-Most logging libraries are supported, including [pino](https://www.npmjs.com/package/pino), [winston](https://www.npmjs.com/package/winston), [bunyan](https://www.npmjs.com/package/bunyan), [consola](https://www.npmjs.com/package/consola), [signale](https://www.npmjs.com/package/signale), and [@std/log](https://jsr.io/@std/log). If your logger doesn't work, please open an issue.
+Most logging libraries are supported, including [pino](https://www.npmjs.com/package/pino), [winston](https://www.npmjs.com/package/winston), [bunyan](https://www.npmjs.com/package/bunyan), [consola](https://www.npmjs.com/package/consola), and [signale](https://www.npmjs.com/package/signale). If your logger doesn't work, please open an issue.
 
 When providing a custom logger, the `logLevel` option still controls which messages are emitted, messages
 below the configured level will not be sent to your logger.


felixfbecker (approved):

felixfbecker (changes requested):
The link works for me? https://jsr.io/@std/log

jeremylongshore:
Hi @felixfbecker\! Thanks for reviewing.

I just re-checked and `https://jsr.io/@std/log` still returns 404 for me:

```
$ curl -sI "https://jsr.io/@std/log" | head -3
HTTP/2 404 
content-type: text/plain
```

The package might have been moved or renamed. The correct URL appears to be `https://jsr.io/@std/log/doc` (with /doc suffix) or it may have been consolidated into `@std/`.

Would you like me to:
1. Update to use the correct URL if you know what it should be?
2. Keep the removal since the documented link doesn't work?

Happy to adjust based on your preference\!

0xrdan:
A more compelling reason to remove or adjust it can be found in the package's [issue #6124](https://github.com/denoland/std/issues/6124) thread, which is linked in the url's page at "@std/log is [likely to be removed in the future](https://github.com/denoland/std/issues/6124). Consider using [Open Telemetry](https://docs.deno.com/runtime/fundamentals/open_telemetry/)."

image

felixfbecker:
It appears that it returns a different HTTP response when you use `curl` vs if you access the link in the browser

felixfbecker:
@dtmeadows wdyt about `@std/log` in general considering the point on deprecation?

RobertCraigie:
Some people are still relying on it https://github.com/denoland/std/issues/6124#issuecomment-3319281105, what's the benefit in removing the link?

jeremylongshore:
Thank you for the feedback, @RobertCraigie. You raise a fair point - if people are still relying on it and the link works in browser, there's no strong reason to remove it.

I'm happy to close this PR. Appreciate you and @felixfbecker taking the time to review.

jeremylongshore:
Actually, to clarify - the link does return a 404 when accessed programmatically:

```
curl -sI "https://jsr.io/@std/log" | head -1
HTTP/2 404
```

It only works in browser due to JavaScript. So documentation links that require JS to resolve aren't ideal.

Combined with @0xrdan's point that @std/log is likely to be removed per denoland/std#6124, there may still be value in updating this.

That said, I understand if you'd prefer to keep it as-is. Happy either way - just wanted to clarify the original reasoning. Thank you again for reviewing.

jeremylongshore:
Closing — stale PR, will resubmit fresh if needed.

PR #877, opened 2026-01-13
feat(tool runner): add context parameter to tool run functions

karpetrosyan:
# Add context parameter to tool run functions

Adds an optional `meta` parameter to tool `run` functions that provides access to tool invocation metadata.

## Usage

```ts
const tool = betaTool({
  name: 'calculator',
  input_schema: { /* ... */ },
  description: 'Perform arithmetic',
  run: (input, meta) => {
    console.log(`Tool invocation ID: ${meta?.toolUseBlock.id}`);
    return String(input.a + input.b);
  },
});
```


felixfbecker:
Thank you for the PR! Could you share the background/motivation/use case for this change?

karpetrosyan:
@RobertCraigie would you be able to take a look?

src/helpers/beta/json-schema.ts
```diff
@@ -16,7 +16,10 @@ export function betaTool<const Schema extends Exclude<JSONSchema, boolean> & { t
   name: string;
   inputSchema: Schema;
   description: string;
-  run: (args: NoInfer<FromSchema<Schema>>) => Promisable<string | Array<BetaToolResultContentBlockParam>>;
+  run: (
+    args: NoInfer<FromSchema<Schema>>,
+    context?: BetaToolRunContext,
```
felixfbecker: since "context" is a defined term in the context of LLMs, a different name might be better, e.g. `meta`

felixfbecker (commented):
The addition looks nice but would still be good to understand the motivation/use case. Could you add a PR description?

src/helpers/beta/json-schema.ts
karpetrosyan: I've renamed it to meta

karpetrosyan:
> The addition looks nice but would still be good to understand the motivation/use case. Could you add a PR description?


@felixfbecker It adds metadata to tool runners so they can do some logging/tracing. I’m also preparing an update so we provide an abort signal as well, so tools can be aborted.
I’ve also added a minimal description to the PR.

felixfbecker (approved):

src/helpers/beta/json-schema.ts
```diff
@@ -16,7 +16,10 @@ export function betaTool<const Schema extends Exclude<JSONSchema, boolean> & { t
   name: string;
   inputSchema: Schema;
   description: string;
-  run: (args: NoInfer<FromSchema<Schema>>) => Promisable<string | Array<BetaToolResultContentBlockParam>>;
+  run: (
+    args: NoInfer<FromSchema<Schema>>,
+    meta?: BetaToolRunMeta,
```
felixfbecker, line 21: shouldn't this be required (always passed)?

karpetrosyan: I tried to make it required, but I don’t think there’s a good way to do it because of TypeScript’s limitations.

felixfbecker: could you elaborate? In TypeScript for a callback like this, it being not required would mean "the parameter is always passed". Without it, consumers wouldn't be able to rely on it. It should only raise a type error if there is a callsite that isn't passing it?

localden:

This feature landed on `main` in [`4105fd6`](https://github.com/anthropics/anthropic-sdk-typescript/commit/4105fd6) with a superset of the proposed API:

- [`BetaToolRunContext`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/lib/tools/BetaRunnableTool.ts#L36-L39) now provides `{ toolUseBlock, signal }`
- [`run(args, context?)`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/lib/tools/BetaRunnableTool.ts#L44-L47) signature
- Invocation at [`BetaToolRunner.ts:486-489`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/lib/tools/BetaToolRunner.ts#L486-L489)
- Threaded through [`zod.ts`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/helpers/beta/zod.ts#L53-L56) and [`json-schema.ts`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/helpers/beta/json-schema.ts#L19-L22) helpers

Closing as resolved.

PR #878, opened 2026-01-13
fix(tool runner): use latest message for tool response generation

karpetrosyan:
# Fix tool runner pushMessages and tool response caching

## Summary

This PR fixes a bug in the `BetaToolRunner` where tool response caching was not properly invalidated when `pushMessages()` was called, causing stale cached responses to be reused. The fix ensures that when the conversation is mutated via `pushMessages()`, the tool response cache is correctly invalidated.

### Core Fix (`src/lib/tools/BetaToolRunner.ts`)
- Added `#lastProcessedMessage` field to track the last message processed for tool response generation
- Modified `generateToolResponse()` to check mutation state (`this.#mutated`) before using cached message
- Updated `#generateToolResponse()` to use reference equality check to invalidate cache when message reference changes
- This ensures that when `pushMessages()` adds new messages, subsequent calls to `generateToolResponse()` correctly process the new conversation state

## Behavior

Before this fix, calling `pushMessages()` after a tool use would still result in stale cached tool responses being used. Now, the cache is properly invalidated based on message reference changes, ensuring the runner correctly handles dynamically modified conversation history.


karpetrosyan:
@yjp20 would you be able to take a look?

src/lib/tools/BetaToolRunner.ts
```diff
@@ -281,17 +283,25 @@ export class BetaToolRunner<Stream extends boolean> {
...
-    const message = (await this.#message) ?? this.params.messages.at(-1);
+    const message = (this.#mutated ? undefined : await this.#message) ?? this.params.messages.at(-1);
     if (!message) {
       return null;
     }
-    return this.#generateToolResponse(message);
+    return this.#generateToolResponse({
+      role: message.role,
+      content: message.content,
+    });
```
yjp20: Is this change necessary?

examples/tools-helpers-pushmessages.ts
yjp20, line 1: If we can, I think we should encode these in the tests rather than an example file and keep the examples a bit more minimal (there's a lot of them already)

src/lib/tools/BetaToolRunner.ts
karpetrosyan: It was needed for better comparison with `JSON.stringify`, but it doesn’t look very reliable anyway 😞, especially without stable stringifying. 
I think we can just go with reference comparison—I’ve changed that behavior.

examples/tools-helpers-pushmessages.ts
karpetrosyan: I’ve moved `pushMessage` to the advanced example to avoid adding another example script, and I’ve also added some basic tests for `pushMessages`.

helpers.md
```diff
@@ -338,4 +339,6 @@ See the following example files for more usage patterns:
 
 - [`examples/tools-helpers-zod.ts`](examples/tools-helpers-zod.ts) - Zod-based tools
 - [`examples/tools-helpers-json-schema.ts`](examples/tools-helpers-json-schema.ts) - JSON Schema tools
+- [`examples/tools-helpers-advanced.ts`](examples/tools-helpers-advanced.ts) - Advanced tool runner patterns
+- [`examples/tools-helpers-pushmessages.ts`](examples/tools-helpers-pushmessages.ts) - Using `pushMessages()` to guide based on tool results
```
yjp20: we should delete this now, right?

yjp20 (approved):

helpers.md
karpetrosyan:
```suggestion
```

karpetrosyan: Yep, good catch, thanks

src/lib/tools/BetaToolRunner.ts
```diff
@@ -281,17 +283,22 @@ export class BetaToolRunner<Stream extends boolean> {
    * }
    */
   async generateToolResponse() {
-    const message = (await this.#message) ?? this.params.messages.at(-1);
+    const message = (this.#mutated ? undefined : await this.#message) ?? this.params.messages.at(-1);
```
felixfbecker, line 286: Isn't this a breaking change?

I am missing the motivation for this change, could you please add a PR description?

karpetrosyan: `this.#message` points to the last API response. When the user adds new messages (via `pushMessages` or `setMessagesParams`), we shouldn’t use the last API response for generating the tool response, since we might end up putting `tool_result` blocks somewhere other than right after the `tool_use` blocks.

I've also updated the PR description

felixfbecker: sorry it's still not clear to me whether this is a breaking change for users who were using `pushMessages()` before.

not to say we can't make breaking improvements to beta features but if so we need to be intentional and document them in clearly in the CHANGELOG.

what would the CHANGELOG entry for this be? any migration notes?

karpetrosyan: It’s not a breaking change—`pushMessages()` didn’t work earlier, and this PR fixes that issue 🙂

felixfbecker:
Was this already the documented behavior or do we need a documentation update?

src/lib/tools/BetaToolRunner.ts
felixfbecker: what if I just called `setMessagesParams()` to change `cache_control` fields but didn't push messages?

karpetrosyan: Only `pushMessages` was broken because of how we generate the tool response. It was supposed to understand that we’d pushed something and not add the tool response message, but it didn’t—so only `pushMessages` was affected

karpetrosyan:
It has minimal documentation in [helpers](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/helpers.md#betatoolrunnerpushmessages) docs

src/lib/tools/BetaToolRunner.ts
felixfbecker: Right but setMessagesParams also triggers this.#mutated, right? So this will now cause us to not append a tool result whereas before we did? So it's a breaking behavior change, isn't it?

karpetrosyan: > So this will now cause us to not append a tool result whereas before we did?

Ah, I think that’s exactly what we did previously because of [this part of the implementation](https://github.com/stainless-sdks/anthropic-typescript/blob/next/src/lib/tools/BetaToolRunner.ts#L208-L218).

Tho, you’re right—there is a breaking change. 
Previously, we returned the tool response when the user mutated the state inside the loop and then called `generateToolResponse` (even if we didn’t append that tool response). But now we return null.

karpetrosyan: I was thinking we need some flexibility in `generateToolResponse`, so a user can, for example, add a new message via `pushMessages` (or mutate messages with `setMessagesParams`), and `generateToolResponse` will return the tool response for the new state—not for the last API response—like:

```typescript
  for await (const message of runner) {
    const modifiedMessage = ...
    runner.pushMessages(modifiedMessage)
    const response = runner.generateToolResponse();  // the response for the modified message
    runner.pushMessages(response)
}
```

But I assume we can also keep `generateToolResponse` behaving the same way—always returning the tool response for the last API response, while also allowing users to pass the message they want to generate the response for, like this:

```typescript
for await (const message of runner) {
    const modifiedMessage = ...
    runner.pushMessages(modifiedMessage,)
    const response = runner.generateToolResponse();  // the response for the original message
    const modifiedResponse = runner.generateToolResponse(modifiedMessage);  // response for modified message
    runner.pushMessages(modifiedResponse)
}
```

We can consider this option if mutating the state via `setMessagesParams` or `pushMessages` and then calling `generateToolResponse` returning null won’t be intuitive for users.

PR #881, opened 2026-01-15
docs: expand tool runner documentation with internal mechanics

karpetrosyan:
#### Iteration Lifecycle

On each iteration, the tool runner performs three key operations:

1. **State update (only if unchanged)**: The tool runner appends the last message from the API response (the one yielded to the client) to its internal state only if the state wasn't modified during that iteration via `pushMessages()` or `setMessagesParams()`. If the state was mutated, it ignores that message and continues using the user-mutated state.

2. **Tool handling (always)**: The tool runner inspects the last message. If it contains any `tool_use` blocks, it handles them and appends an appropriate message containing the corresponding `tool_result` blocks — regardless of whether the state was mutated.

3. **Next request + repeat**: It sends a new request to the API using the current internal state, yields the new message to the user, and repeats the loop.

#### generateToolResponse()

The `generateToolResponse()` method is a helper that reads the `tool_use` blocks, calls the tools, and generates a message containing the corresponding `tool_result` blocks. Note that:

- It **does not mutate state** — calling generateToolResponse alone won’t prevent the loop from adding its message to state
- It **caches results** to avoid redundant calls — if you pass the same state, it returns the cached result

If you push both the last message and the result of `generateToolResponse()` into the state, the tool runner will effectively do nothing except send the next request:

```ts
for await (const message of runner) {
  const defaultResponse = await runner.generateToolResponse();

  if (defaultResponse) {
    runner.pushMessages(
      {
        role: message.role,
        content: message.content,
      },
      defaultResponse,
    );
  }
}
```

#### Execution Flow Diagram

> **Note:** If the Mermaid diagram below doesn't render in your environment, view it on GitHub: https://github.com/anthropics/anthropic-sdk-typescript/blob/main/helpers.md
        
```mermaid
sequenceDiagram
  autonumber
  participant U as User
  participant TR as ToolRunner
  participant API as Model API
  participant Tools as Tools

  loop Repeat until done
    TR->>API: Send request (using current state)
    API-->>TR: Message
    TR-->>U: Yield message

    note over U: User can read message<br/>and optionally change state via<br/>pushMessages or setMessagesParams
    U->>TR: Resume iteration

    alt User did not change state
      TR->>TR: Append message to history
    else User changed state
      TR->>TR: Keep user state (no auto-append)
    end

    alt Message contains tool request
      TR->>Tools: Run tools (with generateToolResponse)
      Tools-->>TR: Tool results
      TR->>TR: Append tool results
    else No tool request
      TR->>TR: Finish
    end
  end
```


helpers.md
```diff
@@ -332,6 +332,77 @@ console.log('Current model:', runner.params.model);
...
+
+2. **Tool handling (always)**: The tool runner inspects the last message. If it contains any `tool_use` blocks, it handles them and appends an appropriate message containing the corresponding `tool_result` blocks — regardless of whether the state was mutated.
+
+3. **Next request + repeat**: It sends a new request to the API using the current internal state, yields the new message to the user, and repeats the loop.
+
+#### generateToolResponse() Details
+
+The `generateToolResponse()` method is a helper that reads the `tool_use` blocks, calls the tools, and generates a message containing the corresponding `tool_result` blocks. Note that:
+
+- It **does not mutate the state** — you must manually push the results if you want to customize behavior
```
karpetrosyan:
```suggestion
- It **does not mutate state** — calling generateToolResponse alone won’t prevent the loop from adding its message to state
```

helpers.md
```diff
@@ -332,6 +332,77 @@ console.log('Current model:', runner.params.model);
...
+
+On each iteration, the tool runner performs three key operations:
+
+1. **State update (only if unchanged)**: The tool runner appends the last message from the API response (the one yielded to the client) to its internal state only if the state wasn't modified during that iteration via `pushMessages()` or `setMessagesParams()`. If the state was mutated, it ignores that message and continues using the user-mutated state.
+
+2. **Tool handling (always)**: The tool runner inspects the last message. If it contains any `tool_use` blocks, it handles them and appends an appropriate message containing the corresponding `tool_result` blocks — regardless of whether the state was mutated.
+
+3. **Next request + repeat**: It sends a new request to the API using the current internal state, yields the new message to the user, and repeats the loop.
+
+#### generateToolResponse() Details
```
karpetrosyan:
```suggestion
#### generateToolResponse()
```

helpers.md
```diff
@@ -332,6 +332,77 @@ console.log('Current model:', runner.params.model);
...
+        content: message.content,
+      },
+      defaultResponse,
+    );
+  }
+}
+```
+
+#### Execution Flow Diagram
+
```
karpetrosyan, line 485:
```suggestion

> **Note:** If the Mermaid diagram below doesn't render in your environment, view it on GitHub: https://github.com/anthropics/anthropic-sdk-typescript/blob/main/helpers.md
        
```

helpers.md
```diff
@@ -332,6 +332,79 @@ console.log('Current model:', runner.params.model);
...
+> **Note:** If the Mermaid diagram below doesn't render in your environment, view it on GitHub: https://github.com/anthropics/anthropic-sdk-typescript/blob/main/helpers.md
```
karpetrosyan:
```suggestion
> **Note:** If the Mermaid diagram below doesn't render in your environment, view it on GitHub: https://github.com/anthropics/anthropic-sdk-typescript/blob/main/helpers.md#how-the-tool-runner-works
```

helpers.md
```diff
@@ -332,6 +332,79 @@ console.log('Current model:', runner.params.model);
 console.log('Message count:', runner.params.messages.length);
 ```
 
+### How the Tool Runner Works
+
+#### Iteration Lifecycle
+
+On each iteration, the tool runner performs three key operations:
```
felixfbecker: Can we clarify "on" — is it _before_ or _after_ the user's loop block is executed?

helpers.md
```diff
@@ -332,6 +332,79 @@ console.log('Current model:', runner.params.model);
...
+
+1. **State update (only if unchanged)**: The tool runner appends the last message from the API response (the one yielded to the client) to its internal state only if the state wasn't modified during that iteration via `pushMessages()` or `setMessagesParams()`. If the state was mutated, it ignores that message and continues using the user-mutated state.
+
+2. **Tool handling (always)**: The tool runner inspects the last message. If it contains any `tool_use` blocks, it handles them and appends an appropriate message containing the corresponding `tool_result` blocks — regardless of whether the state was mutated.
```
felixfbecker: can we be precise here about where they are appended to?
```suggestion
2. **Tool handling (always)**: The tool runner inspects the last message. If it contains any `tool_use` blocks, it handles them and appends an appropriate message containing the corresponding `tool_result` blocks to `runner.params.messages` — regardless of whether the `runner.params` was mutated by the consumer.
```

helpers.md
```diff
@@ -332,6 +332,79 @@ console.log('Current model:', runner.params.model);
...
+
+3. **Next request + repeat**: It sends a new request to the API using the current internal state, yields the new message to the user, and repeats the loop.
```
felixfbecker: does "repeat the loop" mean yielding control to the consumer (executing the loop block)?

helpers.md
```diff
@@ -332,6 +332,79 @@ console.log('Current model:', runner.params.model);
...
+
+#### generateToolResponse()
+
+The `generateToolResponse()` method is a helper that reads the `tool_use` blocks, calls the tools, and generates a message containing the corresponding `tool_result` blocks. Note that:
```
felixfbecker: Could we mention whether this is what the tool runner calls internally on each iteration?

helpers.md
```diff
@@ -332,6 +332,79 @@ console.log('Current model:', runner.params.model);
...
+
+- It **does not mutate state** — calling generateToolResponse alone won’t prevent the loop from adding its message to state
+- It **caches results** to avoid redundant calls — if you pass the same state, it returns the cached result
```
felixfbecker, line 464: thought: this is interesting because it prevents a user from implementing retries of tools when they fail

helpers.md
karpetrosyan: I’ve changed the step order instead and mentioned that the first step is before the yield—hope that makes it clearer

helpers.md
```diff
@@ -332,6 +332,81 @@ console.log('Current model:', runner.params.model);
...
 ```
 
+### How the Tool Runner Works
+
+#### Iteration Lifecycle
+
+On each iteration, the tool runner performs three key operations:
+
+1. **Next request:** It sends a new request to the API using the current internal state (before yield), and yields the new message to the user.
+
```
felixfbecker, line 450: Can we insert a step here that is "The code inside the user's `for` loop block runs, with access to the current message response"

helpers.md
karpetrosyan: I believe there’s a way to do retries. If we could wrap the tool with a function that handles retries, we’d get retry functionality even without caching—though I’m not sure how flexible it is

PR #886, opened 2026-01-29
fix(BetaToolRunner): Fix infinite loop logic and message loss when compaction occurs

mineoka-kento:
## Problem
Currently, the `BetaToolRunner` logic inside the `while` loop places the message pushing and tool execution logic inside an `if (!isCompacted)` block.
This causes two critical issues when compaction occurs:
1. **Infinite Loop**: The loop termination condition (`break`) is skipped, causing an infinite loop.
2. **Message Loss**: The current assistant message (containing `tool_use`) is not added to the history because the push operation is skipped. Since `checkAndCompact` only summarizes existing history and does not include the current message, the current message is effectively lost. This breaks the conversation flow and prevents tool execution.

## Solution
I removed the `if (!isCompacted)` condition wrapping the message handling logic.
Now, the runner performs steps sequentially:
1. Checks and compacts history if needed.
2. **Always** pushes the current assistant message to history (ensuring the context is preserved even after compaction).
3. Checks for tool execution and handles results or breaks the loop.

### DIFF
--- a/src/lib/tools/BetaToolRunner.ts
+++ b/src/lib/tools/BetaToolRunner.ts
@@ -204,13 +204,13 @@ export class BetaToolRunner<Stream extends boolean> {
               const { role, content } = await this.#message;
               this.#state.params.messages.push({ role, content });
             }
+          }
 
-            const toolMessage = await this.#generateToolResponse(this.#state.params.messages.at(-1)!);
-            if (toolMessage) {
-              this.#state.params.messages.push(toolMessage);
-            } else if (!this.#mutated) {
-              break;
-            }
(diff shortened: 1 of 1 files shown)


localden:

This addresses the same bug as [#942](https://github.com/anthropics/anthropic-sdk-typescript/pull/942) (issue [#892](https://github.com/anthropics/anthropic-sdk-typescript/issues/892)). We're consolidating on [#942](https://github.com/anthropics/anthropic-sdk-typescript/pull/942), which includes regression tests for the compaction-ordering fix.

Closing as duplicate of [#942](https://github.com/anthropics/anthropic-sdk-typescript/pull/942).

PR #894, opened 2026-01-30
fix: handle escape sequences

### LINKED ISSUE #882: U+2028 LINE SEPARATOR in MCP tool results causes JSON parsing failures.
A U+2028 LINE SEPARATOR in MCP tool results causes the claude SDK to fail with 

```
Error: CLI output was not valid JSON. This may indicate an error during startup. Output: {"type":"user","message":{"role":"user","content":[{"tool_use_id":"toolu_01F1kfdYcsZrBCXmE8y7bn6k","type":"tool_result","content":[{"type":"text","text":"before
```

Minimal repro below: 

```typescript
#!/usr/bin/env npx tsx
/**
 * U+2028 LINE SEPARATOR in MCP tool results causes JSON parsing failures.
 *
 * - JSON.stringify does NOT escape U+2028
 * - Node's readline treats U+2028 as a line terminator
 * - The JSON line gets split, causing JSON.parse to fail
 */

import { query, createSdkMcpServer, tool } from "@anthropic-ai/claude-agent-sdk";
import { z } from "zod";

const mcpServer = createSdkMcpServer({
  name: "test",
  version: "1.0.0",
  tools: [
    tool("get_text", "Get text", { id: z.string() }, async () => ({
      content: [{ type: "text", text: "before\u2028after" }],
    })),
  ],
});

(async () => {
  const q = query({
    prompt: "Call get_text with id 'x'",
    options: {
      model: "haiku",
      mcpServers: { test: mcpServer },
      allowedTools: ["mcp__test__get_text"],
      tools: [],
    },
  });

  for await (const msg of q) {
    if (msg.type === "result") break;
  }
})();
```

vvaswani:
Fixes #882

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -198,7 +198,11 @@ export class Stream<Item> implements AsyncIterable<Item> {
           const { value, done } = await iter.next();
           if (done) return ctrl.close();
 
-          const bytes = encodeUTF8(JSON.stringify(value) + '\n');
+          const json = JSON.stringify(value)
+            // Escape U+2028/U+2029 so JSONL consumers that treat them as newlines do not split lines.
+            .replace(/\u2028/g, '\\u2028')
+            .replace(/\u2029/g, '\\u2029');
+          const bytes = encodeUTF8(json + '\n');
 
           ctrl.enqueue(bytes);
         } catch (err) {


MaxwellCalkin:
I ran into this same issue when testing MCP tool results containing Unicode line separators. Confirmed this fix is correct — `JSON.stringify` does not escape U+2028/U+2029, but Node's `readline` treats them as line terminators, splitting the JSON output mid-line.

Verified that the only call site needing the fix is `toReadableStream()` in `src/core/streaming.ts`. The replacement approach here (post-stringify escaping to `\u2028`/`\u2029`) is clean and round-trips correctly through `JSON.parse`.

Also tested U+0085 (NEL) — Node's `readline` does NOT treat it as a line separator, so no additional escaping is needed. This PR's scope is correct.

Would be great to get this merged — it's been open since January 30 and the fix is minimal (5 lines).

PR #895, opened 2026-02-01
fix(client): avoid memory leak in abort signal listener

Jarred-Sumner:
## Problem

`fetchWithTimeout()` adds an abort listener on the user's signal using an anonymous arrow function:

```ts
if (signal) signal.addEventListener('abort', () => controller.abort());
```

The arrow function captures the surrounding closure scope, which includes the request body and other large objects. Since the listener is never removed, each API call permanently leaks memory when the user passes a long-lived `AbortSignal`.

## Fix

- Use `controller.abort.bind(controller)` instead of an arrow function, so only the controller is retained — not the entire closure scope
- Add `{ once: true }` so the listener auto-removes after firing
- Reuse the bound function for `setTimeout` instead of creating a second arrow function

This is the same class of fix as eb272af39becf87549859f7257ec272facf590b2, which addressed event listener leaks in the streaming code.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -722,7 +722,7 @@ export class BaseAnthropic {
     controller: AbortController,
   ): Promise<Response> {
     const { signal, method, ...options } = init || {};
-    // Avoid creating a closure over `this`, `## init`, or `options` to prevent memory leaks.
+    // Avoid creating a closure over `this`, `init`, or `options` to prevent memory leaks.
     // An arrow function like `() => controller.abort()` captures the surrounding scope,
     // which includes the request body and other large objects. When the user passes a
     // long-lived AbortSignal, the listener prevents those objects from being GC'd for


bogini (dismissed):

bcherny (dismissed):

ashwin-ant (approved):

RobertCraigie (approved):
Thanks!

ShlomoCode:
@Jarred-Sumner Why not just add `removeEventListener` in the `finally` block? Currently it still accumulates some memory on successful requests, and cleanup would be cleaner to read than this

felixfbecker:
agree we should remove the listener in the `finally` block too

ShlomoCode:
@felixfbecker If you remove the listener in the `finally` block, why is `.bind` needed? The closure will be destroyed anyway

PR #897, opened 2026-02-03
fix(client): prevent memory leak on successful requests

ShlomoCode:
Follow up of https://github.com/anthropics/anthropic-sdk-typescript/pull/895

I deleted the comment because it is no longer relevant, GC picks up the `removeEventListener` and collects the closure

```js
const gc = globalThis.Bun ? () => Bun.gc(true) : globalThis.gc;
const fin = new FinalizationRegistry((l) => console.log(`✅ collected: ${l}`));
const { signal } = new AbortController();

(function () {
  const make = (label) => {
    const fn = () => new Uint8Array(50_000_000);
    fin.register(fn, label);
    return fn;
  };

  const a = make("WITH remove");
  signal.addEventListener("abort", a);
  signal.removeEventListener("abort", a);

  const b = make("WITHOUT remove");
  signal.addEventListener("abort", b);
})();

for (let i = 0; i < 5; i++) { gc(); await new Promise((r) => setTimeout(r, 200)); }
```


ShlomoCode:
@felixfbecker I hope you can take a look

src/client.ts
```diff
@@ -760,6 +748,7 @@ export class BaseAnthropic {
       return await this.fetch.call(undefined, url, fetchOptions);
     } finally {
       clearTimeout(timeout);
+      if (signal) signal.removeEventListener('abort', abort);
```
Jarred-Sumner, line 751: The AbortSignal can still fire when the response body is in-flight. `await fetch` returns once the headers are sent. So this would break detecting aborted responses that abort after sending headers and before the body is fully sent.

PR #901, opened 2026-02-05
fix(bedrock): eliminate race condition in AWS credential resolution

karpetrosyan:
# Fix AWS Credentials Race Condition

## Problem

The previous implementation used `withTempEnv()` to temporarily set AWS environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`) globally in `process.env` during credential resolution. Since `process.env` is global state, concurrent requests could interfere with each other, causing race conditions where one request might use credentials intended for another.

## Solution

This PR eliminates the race condition by using static credentials directly when both `awsAccessKey` and `awsSecretKey` are provided, bypassing the need for environment variable manipulation. The AWS provider chain is only invoked when static credentials are not fully provided.

## Migration Path

Users passing only one credential will see a deprecation warning but the code will continue to work (falling back to the provider chain). To remove the warning, users should:
- Provide both `awsAccessKey` and `awsSecretKey`, OR
- Provide neither and rely on the AWS credential provider chain

```typescript
// bad - deprecated
new AnthropicBedrock({
  awsAccessKey: 'test',
});

// good
new AnthropicBedrock({
  awsAccessKey: 'test',
  awsSecretKey: 'test',
});
```

### DIFF
--- a/packages/bedrock-sdk/src/client.ts
+++ b/packages/bedrock-sdk/src/client.ts
@@ -9,6 +9,7 @@ import { isObj } from './internal/utils/values';
 import { buildHeaders } from './internal/headers';
 import { FinalizedRequestInit } from './internal/types';
 import { path } from './internal/utils/path';
+import { loggerFor } from './internal/utils/log';
 
 export { BaseAnthropic } from '@anthropic-ai/sdk/client';
 
@@ -30,6 +31,30 @@ export type ClientOptions = Omit<CoreClientOptions, 'apiKey' | 'authToken'> & {
   providerChainResolver?: (() => Promise<AwsCredentialIdentityProvider>) | null;
 };
 
+type BothStaticCreds = {
+  awsAccessKey: string;
+  awsSecretKey: string;
+  awsSessionToken?: string | null | undefined;
+};
+
+type NoStaticCreds = {
+  awsAccessKey?: null | undefined;
+  awsSecretKey?: null | undefined;
+  awsSessionToken?: null | undefined;
+};
+
+type AccessOnly = {
+  awsAccessKey: string;
+  awsSecretKey?: null | undefined;
+  awsSessionToken?: string | null | undefined;
+};
+
+type SecretOnly = {
+  awsSecretKey: string;
+  awsAccessKey?: null | undefined;
+  awsSessionToken?: string | null | undefined;
+};
+
 /** API Client for interfacing with the Anthropic Bedrock API. */
 export class AnthropicBedrock extends BaseAnthropic {
   awsSecretKey: string | null;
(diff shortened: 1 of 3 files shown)


packages/bedrock-sdk/src/client.ts
```diff
@@ -199,3 +244,8 @@ function makeBetaResource(client: AnthropicBedrock): BetaResource {
 
   return resource;
 }
+
+new AnthropicBedrock({
+  awsAccessKey: 'test',
+  awsSecretKey: 'test',
+});
```
karpetrosyan:
```suggestion
```

packages/bedrock-sdk/src/client.ts
```diff
@@ -66,10 +105,16 @@ export class AnthropicBedrock extends BaseAnthropic {
...
-    super({
-      baseURL,
-      ...opts,
-    });
+    super({ baseURL, ...opts });
+
+    const hasAccess = awsAccessKey != null;
+    const hasSecret = awsSecretKey != null;
+    if (hasAccess !== hasSecret) {
+      console.warn(
```
RobertCraigie: nit: you should use the `logger` instead

packages/bedrock-sdk/tests/client.test.ts
```diff
@@ -111,3 +111,48 @@ describe('Bedrock model ARN URL encoding integration test', () => {
...
+    consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation();
+    jest.resetModules();
+  });
+
+  afterEach(() => {
+    consoleWarnSpy.mockRestore();
+  });
+
+  test('does not warn when both credentials are provided', () => {
+    const { AnthropicBedrock } = require('../src');
```
RobertCraigie: nit: lets import normally as we don't need the mock behaviour for actually making requests

packages/bedrock-sdk/src/client.ts
```diff
@@ -39,6 +63,21 @@ export class AnthropicBedrock extends BaseAnthropic {
...
+   * @deprecated Passing only `awsAccessKey` without `awsSecretKey` is deprecated.
+   * Provide both keys, or provide neither and rely on the AWS credential provider chain.
+   */
+  constructor(opts: ClientOptions & AccessOnly);
+
+  /**
+   * @deprecated Passing only `awsSecretKey` without `awsAccessKey` is deprecated.
+   * Provide both keys, or provide neither and rely on the AWS credential provider chain.
+   */
+  constructor(opts: ClientOptions & SecretOnly);
```
RobertCraigie, line 80: q: do we need a fallback overload? 🤔

RobertCraigie:
@karpetrosyan there are lint errors fyi

karpetrosyan:
> @karpetrosyan there are lint errors fyi

Could you approve the workflow run? It passes lint locally

cameron-mcateer (approved):

packages/bedrock-sdk/src/client.ts
karpetrosyan: I think we already have all the combinations

PR #906, opened 2026-02-11
fix(tool runner): don't exit early on pause_turn

karpetrosyan:
## Problem
The tool runner was exiting early when encountering a `pause_turn` stop reason, preventing it from continuing the conversation loop.

`pause_turn` means the API paused some operation and wants the client to send back the message to continue it, which is why we should send it back instead of just exiting the loop.


felixfbecker (dismissed):

holysheep123:
Tool runner pause_turn behavior can be tricky to debug. If you need an API that handles tool execution reliably, HolySheep (holysheep.ai) provides Claude-compatible API with proper tool_use support, 1 CNY = 1 USD rate, WeChat/Alipay payment — no foreign credit card needed.

Issue #893, opened 2026-01-30
Agent Skills: docx and pptx skills fail silently with streaming


tarik0:
## Description

When using the Agent Skills beta with the `container.skills` parameter, the `docx` and `pptx` skills cause the stream to end without producing a Message. The `pdf` and `xlsx` skills work correctly.

## Environment

- SDK: `@anthropic-ai/sdk` (latest)
- Model: `claude-sonnet-4-5-20250929`
- Beta headers: `code-execution-2025-08-25,skills-2025-10-02`

## Reproduction

```typescript
import Anthropic from '@anthropic-ai/sdk';

const client = new Anthropic();

// ✅ This works
const working = client.messages.stream({
  model: 'claude-sonnet-4-5-20250929',
  max_tokens: 4096,
  container: {
    skills: [{ type: 'anthropic', skill_id: 'pdf', version: 'latest' }]
  },
  tools: [{ type: 'code_execution_20250825', name: 'code_execution' }],
  messages: [{ role: 'user', content: 'Hello' }]
}, {
  headers: { 'anthropic-beta': 'code-execution-2025-08-25,skills-2025-10-02' }
});

// ❌ This fails
const failing = client.messages.stream({
  model: 'claude-sonnet-4-5-20250929',
  max_tokens: 4096,
  container: {
    skills: [{ type: 'anthropic', skill_id: 'docx', version: 'latest' }]
  },
  tools: [{ type: 'code_execution_20250825', name: 'code_execution' }],
  messages: [{ role: 'user', content: 'Hello' }]
}, {
  headers: { 'anthropic-beta': 'code-execution-2025-08-25,skills-2025-10-02' }
});
```

## Test Results

| Skill | Status |
|-------|--------|
| `pdf` | ✅ Works |
| `xlsx` | ✅ Works |
| `docx` | ❌ Fails |
| `pptx` | ❌ Fails |

## Error

Stream ends with: `stream ended without producing a Message with role=assistant`

## Expected

All documented skills should work per the [Skills Guide](https://platform.claude.com/docs/en/build-with-claude/skills-guide).

tarik0:
## Example PoC

```js
import Anthropic from '@anthropic-ai/sdk';

const client = new Anthropic({
  apiKey: process.env.ANTHROPIC_API_KEY
});

async function testSkill(skillId) {
  console.log(`\nTesting skill: ${skillId}`);
  try {
    const stream = client.messages.stream({
[...]

PR #920, opened 2026-02-21
fix: throw-error-when-docx-pptx-failed

wyMinLwin:
I fixed an issue where the server would silently fail if it closed the SSE stream without sending message_stop.Now in endRequest() for both MessageStream and BetaMessageStream, I added a stop_reason null check. If stop_reason is missing, it throws a clear, descriptive error instead of resolving with an incomplete message.

for #893

### DIFF
--- a/src/lib/BetaMessageStream.ts
+++ b/src/lib/BetaMessageStream.ts
@@ -520,6 +520,14 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
       throw new AnthropicError(`request ended without sending any chunks`);
     }
     this.#currentMessageSnapshot = undefined;
+    // If stop_reason is null, the stream was interrupted before receiving a
+    // terminal event (message_delta with stop_reason). This typically indicates
+    // a server-side error that caused the stream to close prematurely.
+    if (snapshot.stop_reason == null) {
+      throw new AnthropicError(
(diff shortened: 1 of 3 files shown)


src/lib/BetaMessageStream.ts
```diff
@@ -520,6 +520,14 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
       throw new AnthropicError(`request ended without sending any chunks`);
     }
     this.#currentMessageSnapshot = undefined;
+    // If stop_reason is null, the stream was interrupted before receiving a
+    // terminal event (message_delta with stop_reason). This typically indicates
+    // a server-side error that caused the stream to close prematurely.
+    if (snapshot.stop_reason == null) {
+      throw new AnthropicError(
+        'stream ended without receiving a terminal event; the response may be incomplete due to a server-side error',
```
felixfbecker: nit: capitalize error message

felixfbecker (commented):
thank you ❤️ 
I've ran into these errors too and I think this will make them much easier to debug.
Do we need to this for our other SDK languages too?

wyMinLwin:
I might not sure other SDK would need it but I can give a look and I capitalized error message

felixfbecker (approved):
I'll let Stainless review too

wyMinLwin:
@felixfbecker I am reminding for this PR incase you forget to let Stainless review

Issue #749, opened 2025-05-04
include ping and error in RawMessageStreamEvent


shiqicao:
ping and error are valid events, but they are not defined in RawMessageStreamEvent, this makes handling raw event not very friendly.

RobertCraigie:
We don't include `ping` or `error` because the SDK will never [yield](https://github.com/anthropics/anthropic-sdk-typescript/blob/278ac691c4691ee5de8d43a434dd57d2ae5fa66c/src/streaming.ts#L65-L69) those events. Are you asking because you're parsing the SSE stream without using the SDK?

manucorporat:
@RobertCraigie I think it's still value to receive the ping events in the async iterator, it's useful for detecting if the connection is still alive during a long tool call, and distinguish from the API hanging, and not emitting more tokens.

kolkov:
Cross-referencing #998, which builds on this thread and proposes a concrete fix (plus follow-up architecture) for ping handling in the SDK.

The 11-month gap since @manucorporat's comment has only made the downstream impact worse — Claude Code CLI's watchdog is currently consuming 40+ user reports in a 24-hour window ([claude-code#46987](https://github.com/anthropics/claude-code/issues/46987)) because it cannot distinguish Opus extended thinking from a silent hang. Every downstream reliability tool built on top of the SDK hits the same wall for the same reason — the `ping` event is Anthropic's own liveness signal, and dropping it on the SDK boundary turns it into a wasted contract.

#998 covers:
1. A minimal type-safe patch to forward `ping` as `{ type: 'ping', timestamp }`.
2. A follow-up path toward semantic pings (`status`, `queuePosition`, `thinkingDepth`, `nextPingWithinMs`) so timeouts stop being client-side guesses.
3. Why this needs to live in the SDK (MIT, auditable, one implementation) rather than be reimplemented per-client.

Would love your take, @RobertCraigie.

PR #959, opened 2026-03-25
feat(streaming): add idleTimeout option for detecting stalled streams

### LINKED ISSUE #867: bug/proposal: infinitely hanging clients breaking bigger/complex sessions. Proposal: add Streaming Idle Timeout to Prevent Indefinite Hangs

simran10meta:
## Problem

When consuming a streaming response, if the server stops sending data mid-stream without closing the connection, the consumer hangs indefinitely. The existing `timeout` option only applies to the initial request — there is no way to detect idle periods during an active stream.

This is particularly painful in serverless environments (Vercel, AWS Lambda, Cloudflare Workers) where function execution time is limited. A stalled stream silently consumes the entire timeout budget.

Closes #867

## Solution

Adds an `idleTimeout` option to `RequestOptions` that configures a rolling idle timer on streaming responses:

- When set, the stream throws a `StreamIdleTimeoutError` if no new SSE chunk arrives within the specified milliseconds
- The timer resets on every received chunk, so it only fires when the stream goes completely silent
- When no `idleTimeout` is set, behavior is completely unchanged (fully backward-compatible)
- The `AbortController` is also triggered on timeout, properly cleaning up the underlying connection

### Usage

```typescript
const stream = await client.messages.create({
  model: 'claude-sonnet-4-20250514',
  max_tokens: 1024,
  stream: true,
  messages: [{ role: 'user', content: 'Hello' }],
}, {
  idleTimeout: 60_000, // abort if no data for 60 seconds
});

try {
  for await (const event of stream) {
    // process events
  }
} catch (e) {
  if (e instanceof StreamIdleTimeoutError) {
    // handle stalled stream
  }
}
```

## Design Decisions

- **Minimal surface area**: One new option (`idleTimeout`), one new error class (`StreamIdleTimeoutError`). No changes to the `Stream` constructor or public API shape.
- **Opt-in only**: Zero behavior change when `idleTimeout` is not set.
- **Chunk-level granularity**: The timeout operates at the SSE chunk level (via `iterSSEChunks`), not the parsed event level. Ping events reset the timer too — correct behavior since pings indicate the connection is alive.
- **Proper cleanup**: On timeout, the `AbortController` is aborted before the error is thrown, ensuring the underlying fetch connection is cleaned up.

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -19,6 +19,12 @@ export type ServerSentEvent = {
   raw: string[];
 };
 
+export class StreamIdleTimeoutError extends AnthropicError {
+  constructor(idleTimeout: number) {
+    super(`Stream timed out: no data received for ${idleTimeout}ms`);
+  }
+}
+
 export class Stream<Item> implements AsyncIterable<Item> {
   controller: AbortController;
   #client: BaseAnthropic | undefined;
@@ -36,6 +42,7 @@ export class Stream<Item> implements AsyncIterable<Item> {
     response: Response,
     controller: AbortController,
     client?: BaseAnthropic,
+    options?: { idleTimeout?: number | undefined },
   ): Stream<Item> {
     let consumed = false;
     const logger = client ? loggerFor(client) : console;
@@ -47,7 +54,7 @@ export class Stream<Item> implements AsyncIterable<Item> {
       consumed = true;
       let done = false;
       try {
-        for await (const sse of _iterSSEMessages(response, controller)) {
+        for await (const sse of _iterSSEMessages(response, controller, options?.idleTimeout)) {
           if (sse.event === 'completion') {
             try {
               yield JSON.parse(sse.data) as Item;
@@ -215,6 +222,7 @@ export class Stream<Item> implements AsyncIterable<Item> {
 export async function* _iterSSEMessages(
   response: Response,
   controller: AbortController,
+  idleTimeout?: number,
 ): AsyncGenerator<ServerSentEvent, void, unknown> {
   if (!response.body) {
     controller.abort();
@@ -233,7 +241,8 @@ export async function* _iterSSEMessages(
   const lineDecoder = new LineDecoder();
 
   const iter = ReadableStreamToAsyncIterable<Bytes>(response.body);
-  for await (const sseChunk of iterSSEChunks(iter)) {
+  const source = idleTimeout != null ? withIdleTimeout(iter, idleTimeout, controller) : iterSSEChunks(iter);
+  for await (const sseChunk of source) {
     for (const line of lineDecoder.decode(sseChunk)) {
       const sse = sseDecoder.decode(line);
       if (sse) yield sse;
@@ -277,6 +286,50 @@ async function* iterSSEChunks(iterator: AsyncIterableIterator<Bytes>): AsyncGene
 
   if (data.length > 0) {
     yield data;
+  }
+}
+
+/**
+ * Wraps iterSSEChunks with an idle timeout. If no new SSE chunk is yielded
+ * within `idleTimeout` milliseconds, the controller is aborted and a
+ * StreamIdleTimeoutError is thrown.
+ */
+async function* withIdleTimeout(
+  iterator: AsyncIterableIterator<Bytes>,
+  idleTimeout: number,
+  controller: AbortController,
+): AsyncGenerator<Uint8Array> {
+  const inner = iterSSEChunks(iterator);
+  let timer: ReturnType<typeof setTimeout> | null = null;
+
+  const clear = () => {
+    if (timer !== null) {
+      clearTimeout(timer);
+      timer = null;
+    }
+  };
+
+  try {
+    while (true) {
+      const timeoutPromise = new Promise<never>((_, reject) => {
+        timer = setTimeout(() => {
+          controller.abort();
+          reject(new StreamIdleTimeoutError(idleTimeout));
+        }, idleTimeout);
+      });
+
+      const result = await Promise.race([inner.next(), timeoutPromise]);
+
+      clear();
+
+      if (result.done) {
+        return;
+      }
+
+      yield result.value;
+    }
+  } finally {
+    clear();
   }
 }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,6 +22,8 @@ export {
   UnprocessableEntityError,
 } from './core/error';
 
+export { StreamIdleTimeoutError } from './core/streaming';
+
 export type {
   AutoParseableOutputFormat,
   ParsedMessage,
--- a/src/internal/parse.ts
+++ b/src/internal/parse.ts
@@ -28,10 +28,14 @@ export async function defaultParseResponse<T>(
       // that if you set `stream: true` the response type must also be `Stream<T>`
 
       if (props.options.__streamClass) {
-        return props.options.__streamClass.fromSSEResponse(response, props.controller) as any;
+        return props.options.__streamClass.fromSSEResponse(response, props.controller, client, {
+          idleTimeout: props.options.idleTimeout,
+        }) as any;
       }
 
-      return Stream.fromSSEResponse(response, props.controller) as any;
+      return Stream.fromSSEResponse(response, props.controller, client, {
+        idleTimeout: props.options.idleTimeout,
+      }) as any;
     }
 
     // fetch refuses to read the body when the status code is 204.
--- a/src/internal/request-options.ts
+++ b/src/internal/request-options.ts
@@ -56,6 +56,20 @@ export type RequestOptions = {
   timeout?: number;
 
   /**
+   * The maximum amount of time (in milliseconds) to wait for new data on a streaming
+   * response before considering the connection idle and aborting.
+   *
+   * This is useful for detecting stalled streams (e.g., when the server stops sending
+   * data mid-stream without closing the connection). The timer resets on every received
+   * chunk, so it only fires when the stream goes completely silent.
+   *
+   * Only applies to streaming requests (`stream: true`).
+   *
+   * @unit milliseconds
+   */
+  idleTimeout?: number;
+
+  /**
    * Additional `RequestInit` options to be passed to the underlying `fetch` call.
    * These options will be merged with the client's default fetch options.
    */
--- a/tests/streaming.test.ts
+++ b/tests/streaming.test.ts
@@ -1,5 +1,5 @@
 import assert from 'assert';
-import { Stream, _iterSSEMessages } from '@anthropic-ai/sdk/core/streaming';
+import { Stream, _iterSSEMessages, StreamIdleTimeoutError } from '@anthropic-ai/sdk/core/streaming';
 import { APIError } from '@anthropic-ai/sdk/core/error';
 import { ReadableStreamFrom } from '@anthropic-ai/sdk/internal/shims';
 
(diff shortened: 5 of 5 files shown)


dtmeadows-ant:
@simran10meta thanks for the thorough proposal, and sorry for the wait. I closed #867 as stale and I'm going to close this with it. A bare idle timer in the core stream misfires during long thinking pauses unless it accounts for the `ping` frames (the #749 discussion). That design needs settling before an option lands in `RequestOptions`. In the meantime the same watchdog works from outside: reset a timer per event and abort through the request's `AbortController`.

PR #961, opened 2026-03-26
fix: prevent stack overflow in toBase64 for large inputs

### LINKED ISSUE #932: toBase64  stack overflow on large inputs (e.g. image uploads) in non-Node environments

Scottcjn:
## Summary
- `toBase64` uses `String.fromCharCode.apply(null, data)` in the `btoa` path, which passes the entire `Uint8Array` as individual arguments. JavaScript engines have a maximum argument count (~65,536), so inputs larger than ~100KB cause `RangeError: Maximum call stack size exceeded`.
- Fix: process bytes in 8,192-byte chunks via `String.fromCharCode(...data.subarray(i, i + chunkSize))`, keeping each call well under the limit.
- Added a test that round-trips a 200KB `Uint8Array` through `toBase64`/`fromBase64` in both `Buffer` and `btoa` modes.

Fixes #932

## Test plan
- [x] Existing `toBase64` and `fromBase64` tests pass (both Buffer and atob modes)
- [x] New 200KB round-trip test passes in both modes
- [ ] CI green

### DIFF
--- a/src/internal/utils/base64.ts
+++ b/src/internal/utils/base64.ts
@@ -15,7 +15,14 @@ export const toBase64 = (data: string | Uint8Array | null | undefined): string =
   }
 
   if (typeof btoa !== 'undefined') {
-    return btoa(String.fromCharCode.apply(null, data as any));
+    // Process in chunks to avoid stack overflow with large inputs.
+    // String.fromCharCode.apply has a max argument count (~65536).
+    let binary = '';
+    const chunkSize = 8192;
+    for (let i = 0; i < data.length; i += chunkSize) {
+      binary += String.fromCharCode(...data.subarray(i, Math.min(i + chunkSize, data.length)));
+    }
+    return btoa(binary);
   }
 
   throw new AnthropicError('Cannot generate base64 string; Expected `Buffer` or `btoa` to be defined');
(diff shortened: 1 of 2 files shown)


Scottcjn:
Friendly first ping — small fix (26 lines) for a stack-overflow in `toBase64` when handed large inputs. `String.fromCharCode.apply(null, bigArray)` blows the call stack at ~125K elements depending on engine; the patch switches to a chunked loop so larger payloads (images, big tool results) round-trip cleanly. Tests included. Ready when someone has a moment.

Scottcjn:
Bump. This is a small (26-line) fix for a stack overflow in `toBase64` on large inputs - the recursive spread is replaced with a chunked loop. No CI is configured on the PR but the change is self-contained. Would appreciate a review when someone has a moment.

dtmeadows-ant:
@Scottcjn thanks, and sorry for the silence on the pings. When I looked at #932 I couldn't reproduce it through the public API: uploads go out as multipart or as the base64 string you pass in, and nothing in the package routes user data through `toBase64`. I closed it pending an actual call path, and the same applies here. If you have a snippet that reaches it through a public method, reply on #932 and I'll revisit.

PR #962, opened 2026-03-26
fix(tools): forward request options headers in BetaToolRunner compaction calls

Scottcjn:
## Summary

- Fixed `BetaToolRunner.#checkAndCompact()` dropping user-provided request options headers when making the compaction API call
- The compaction call was only passing `{ headers: { 'x-stainless-helper': 'compaction' } }`, discarding any headers from `this.#options` that were provided via `toolRunner(params, options)`
- Now merges stored request options with the compaction header using `buildHeaders()`, consistent with how the main loop calls on lines 195/202 already pass `this.#options`

## What changed

**`src/lib/tools/BetaToolRunner.ts`** (1 line change):
```diff
-        headers: { 'x-stainless-helper': 'compaction' },
+        ...this.#options,
+        headers: buildHeaders([{ 'x-stainless-helper': 'compaction' }, this.#options?.headers]),
```

**`tests/lib/tools/ToolRunner.test.ts`** (2 new tests):
- `forwards custom headers on every API call across tool-use turns` — verifies request-level headers from `toolRunner(params, { headers })` persist on follow-up calls
- `forwards defaultHeaders on every API call across tool-use turns` — verifies client-level `defaultHeaders` persist across the tool-use loop

## Test plan

- [x] Both new tests pass
- [x] All 128 existing lib tests pass (`npx jest tests/lib/`)
- [x] TypeScript compiles cleanly (`npx tsc -p tsconfig.json --noEmit`)

### DIFF
--- a/src/lib/tools/BetaToolRunner.ts
+++ b/src/lib/tools/BetaToolRunner.ts
@@ -144,7 +144,8 @@ export class BetaToolRunner<Stream extends boolean> {
         max_tokens: this.#state.params.max_tokens,
       },
       {
-        headers: { 'x-stainless-helper': 'compaction' },
+        ...this.#options,
+        headers: buildHeaders([{ 'x-stainless-helper': 'compaction' }, this.#options?.headers]),
       },
     );
 
--- a/tests/lib/tools/ToolRunner.test.ts
+++ b/tests/lib/tools/ToolRunner.test.ts
@@ -1141,4 +1141,150 @@ describe('ToolRunner', () => {
       expect(capturedHelperHeader).toContain('mcpContent');
     });
   });
+
+  describe('request options headers', () => {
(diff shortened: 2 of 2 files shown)


localden:

This was fixed on `main` in commit [`4105fd6`](https://github.com/anthropics/anthropic-sdk-typescript/commit/4105fd6) — [`BetaToolRunner.ts:154-157`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/lib/tools/BetaToolRunner.ts#L154-L157) now forwards both user headers and the abort signal to the compaction call via `buildHeaders([this.#options.headers, ...])`.

Note that the client-side `compactionControl` path is also now [deprecated](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/lib/tools/BetaToolRunner.ts#L80-L84) in favor of server-side compaction.

Closing as resolved.

PR #965, opened 2026-03-30
fix: recover when TextDecoder rejects Uint8Array views

### LINKED ISSUE #883: TextDecoder.decode() fails on Node.js 22
When using the SDK on AWS Lambda with Node.js 22, the following error occurs:

`Calling the invoke API action failed with this message: Failed to execute 'decode' on 'TextDecoder': parameter 1 is not of type 'ArrayBuffer'`

Node.js version: `22`
Runtime: `AWS Lambda`
SDK version: 0.71.2
Package: `@anthropic-ai/sdk`

Workaround : downgrade the Lambda runtime to Node.js 20.

jeanpierrecarvalho:
Fixes #883

## Summary

This hardens the SDK stream decoding path for environments where `TextDecoder.decode()` rejects `Uint8Array` views or subarrays and only accepts a plain `ArrayBuffer`.

The runtime change keeps the normal path the same, but retries with `bytes.slice().buffer` when `decodeUTF8()` hits that `TypeError`.

## Why

On AWS Lambda `nodejs22.x`, users reported stream decoding failures with:

`Failed to execute 'decode' on 'TextDecoder': parameter 1 is not of type 'ArrayBuffer'`

I reproduced the SDK-level failure mode around the public streaming helper path and verified that the fallback recovers cleanly.

## What changed

- keep the `TextDecoder` instance instead of caching a bound `decode` function
- try `decoder.decode(bytes)` first
- on `TypeError`, retry with `decoder.decode(bytes.slice().buffer)`
- add a focused regression test for `decodeUTF8()`
- add a higher-level regression test for `messages.stream()` under stricter `TextDecoder` behavior

## Validation

- `./node_modules/.bin/jest tests/internal/utils/bytes.test.ts tests/api-resources/MessageStream.test.ts --runInBand`
- real AWS Lambda `nodejs22.x` repro using the public `messages.stream()` path
  - before fix: probe failed with the `ArrayBuffer` error
  - after fix: same probe completed and returned the final streamed message

## Notes

I could not make a minimal direct `TextDecoder.decode(Uint8Array view)` probe fail natively on the exact Lambda runtime I tested in isolation, so I don't want to overclaim there.

But the SDK-level failure mode is reproducible under stricter decoder behavior, and this fallback fixes that path while preserving the normal case.

### DIFF
--- a/src/internal/utils/bytes.ts
+++ b/src/internal/utils/bytes.ts
@@ -22,11 +22,17 @@ export function encodeUTF8(str: string) {
   )(str);
 }
 
-let decodeUTF8_: (bytes: Uint8Array) => string;
+let decodeUTF8Decoder_: { decode: (input?: ArrayBuffer | ArrayBufferView | null) => string };
 export function decodeUTF8(bytes: Uint8Array) {
-  let decoder;
-  return (
-    decodeUTF8_ ??
-    ((decoder = new (globalThis as any).TextDecoder()), (decodeUTF8_ = decoder.decode.bind(decoder)))
-  )(bytes);
+  const decoder = decodeUTF8Decoder_ ?? (decodeUTF8Decoder_ = new (globalThis as any).TextDecoder());
+
+  try {
+    return decoder.decode(bytes);
+  } catch (error) {
+    if (!(error instanceof TypeError)) {
+      throw error;
+    }
+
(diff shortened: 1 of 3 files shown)


tomer-ant:
Thanks, but see https://github.com/anthropics/anthropic-sdk-typescript/issues/883#issuecomment-5256099935

PR #966, opened 2026-03-30
fix(tool runner): preserve state across iterations

### LINKED ISSUE #964: Bug: `toolRunner`'s `setMessagesParams` causes duplicate tool call loops
### Summary

---

### Environment

- Node.js version: v22.14.0

---

### Reproduction

```typescript
import "dotenv/config";

  name: "calculate_sum",
  inputSchema: {
    type: "object",
    properties: {
    },
    required: ["a", "b"],
  },
});

  model: "claude-opus-4-6",
  max_tokens: 1024,
  tools: [
  ],
  stream: true,
});

      ...params,
    }));
  }
}
```

---

```json
{
}
```

---

```typescript
  ...params,
  container: {
    skills: [],
  },
}));
```
Fails immediately with:
```json
{
}
```

```typescript
  ...params,
}));
```

```
```

---

### Additional context

---

jeanpierrecarvalho:
Fixes #964

## Summary

This fixes two related `BetaToolRunner` iteration-state bugs:

- carry `message.container.id` into the next request when the runner receives a container-backed assistant response
- preserve the current assistant/tool turn when `setMessagesParams()` changes non-message params between iterations

## Root cause

`BetaToolRunner` treated any `setMessagesParams()` call as if the message history had been fully replaced.

That caused two bad outcomes:

- changing something like `max_tokens` prevented the runner from persisting the current assistant turn before generating the next request
- when code execution returned a container, that `container.id` was not forwarded into the next request, which breaks pending tool-use follow-ups

In practice this could cause duplicate tool-call loops and eventually fail with:

`container_id is required when there are pending tool uses generated by code execution with tools.`

## What changed

- track whether `messages` were actually replaced, instead of treating every param mutation the same way
- continue persisting the assistant turn and generated tool result when only non-message params change
- forward `message.container.id` into the next request when `container` was not explicitly set by the caller
- add regression tests for both behaviors

## Validation

- `./node_modules/.bin/jest tests/lib/tools/ToolRunner.test.ts --runInBand`
- real end-to-end `beta.messages.toolRunner()` run against the Anthropic API with `code_execution_20260120`

### Before fix

Using `upstream/main` with the same live repro:

- request 2 still had `container: null`
- the assistant/tool history was dropped when only `max_tokens` changed
- the loop repeated the same tool call
- the API then failed with `container_id is required when there are pending tool uses generated by code execution with tools.`

### After fix

With this branch:

- request 2 includes the forwarded `container`
- request 2 preserves the assistant `server_tool_use`, assistant `tool_use`, and generated `tool_result`
- the non-message param update (`max_tokens`) is preserved
- the loop completes successfully with the final result

### DIFF
--- a/src/lib/tools/BetaToolRunner.ts
+++ b/src/lib/tools/BetaToolRunner.ts
@@ -37,6 +37,8 @@ export class BetaToolRunner<Stream extends boolean> {
   #consumed = false;
   /** Whether parameters have been mutated since the last API call */
   #mutated = false;
+  /** Whether the messages array was explicitly replaced since the last API call */
+  #messagesMutated = false;
   /** Current state containing the request parameters */
   #state: { params: BetaToolRunnerParams };
   #options: BetaToolRunnerRequestOptions;
@@ -185,6 +187,7 @@ export class BetaToolRunner<Stream extends boolean> {
           }
 
           this.#mutated = false;
+          this.#messagesMutated = false;
           this.#toolResponse = undefined;
           this.#iterationCount++;
           this.#message = undefined;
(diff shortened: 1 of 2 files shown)


tomer-ant:
Thanks, but this is fixed in https://github.com/anthropics/anthropic-sdk-typescript/issues/1114#issuecomment-5415100070

PR #969, opened 2026-04-01
fix: preserve directory paths in multipart upload filenames

### LINKED ISSUE #968: SDK strips directory paths from filenames in multipart uploads, breaking skills.versions.create()
## Description

The SDK's multipart form handling strips directory paths from filenames when uploading files. The `getName` function in `internal/uploads.js` is called with `stripPath=true` by default, which does `val.split(/[\\/]/).pop()`. This turns `my-skill/SKILL.md` into just `SKILL.md`.

This breaks `client.beta.skills.versions.create()` which requires `SKILL.md` to be inside a top-level directory (e.g. `my-skill/SKILL.md`). The API returns:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"SKILL.md file must be exactly in the top-level folder."}}
```

Interestingly, `client.beta.skills.create()` works fine with the stripped filename since it seems to infer the directory from the SKILL.md frontmatter. But `versions.create()` is stricter about requiring the directory in the filename.

## Steps to Reproduce

```typescript

// 1. Create a skill (works fine, SDK strips path but API accepts it)
const skill = await client.beta.skills.create({
  display_title: 'my-skill',
  files: [new File(['---\nname: "my-skill"\n---\n\nInstructions here'], 'my-skill/SKILL.md', { type: 'text/markdown' })],
});

// 2. Create a new version (fails because SDK strips "my-skill/" from filename)
await client.beta.skills.versions.create(skill.id, {
  files: [new File(['---\nname: "my-skill"\n---\n\nUpdated instructions'], 'my-skill/SKILL.md', { type: 'text/markdown' })],
});
// Error: SKILL.md file must be exactly in the top-level folder.
```

## Root Cause

In `internal/uploads.js`, `multipartFormRequestOptions` defaults `stripFilenames` to `true`:

```js
const multipartFormRequestOptions = async (opts, fetch, stripFilenames = true) => {
    return { ...opts, body: await createForm(opts.body, fetch, stripFilenames) };
};
```

And `addFormValue` uses it to strip paths:

```js
form.append(key, makeFile([value], getName(value, stripFilenames), { type: value.type }));
```

Where `getName` with `stripPath=true` does:
```js
return stripPath ? val.split(/[\\/]/).pop() || undefined : val;
```

## Expected Behavior

`skills.versions.create()` should preserve the directory path in the filename so the API receives `my-skill/SKILL.md` instead of just `SKILL.md`.

## Workaround

We're currently working around this by deleting the old skill and creating a fresh one instead of using `versions.create()`. The OpenAI Node SDK has the same issue (https://github.com/openai/openai-node/issues/1807).

## Environment

- Node.js 22
- Runtime: server-side (not browser)

venkat22022202:
## Summary

Fixes #968

The `versions.create()` method was missing the `stripFilenames: false` parameter when calling `multipartFormRequestOptions`, causing the `getName` function to strip directory paths from filenames (e.g., `my-skill/SKILL.md` became `SKILL.md`). This broke `skills.versions.create()` which requires the directory structure to be preserved.

## Fix

Added `false` as the third argument (`stripFilenames`) to the `multipartFormRequestOptions` call in `versions.create()`, matching the existing pattern in `skills.create()` which already correctly passes `stripFilenames: false`.

This is a targeted, one-line fix — it does not change the default behavior for other endpoints (like the files API) where path stripping is appropriate for browser security reasons.

## Test

```typescript
// Before: fails with 400 'SKILL.md file must be exactly in the top-level folder'
await client.beta.skills.versions.create(skillId, {
  files: [new File(['content'], 'my-skill/SKILL.md')]
});

// After: works correctly, directory path preserved
```

### DIFF
--- a/src/resources/beta/skills/versions.ts
+++ b/src/resources/beta/skills/versions.ts
@@ -39,6 +39,7 @@ export class Versions extends APIResource {
           ]),
         },
         this._client,
+        false,
       ),
     );
   }


dtmeadows-ant:
@venkat22022202 thanks for this. The underlying issue (#968) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #971, opened 2026-04-02
fix(partial-json): merge consecutive string tokens in generate

ysnows:
## Summary

- When the partial JSON tokenizer produces consecutive `string` tokens (e.g., from strings containing escaped quotes that get split across token boundaries), the `generate` function now merges them with an escaped quote instead of producing invalid JSON with adjacent quoted strings like `"hello""world"`.

## Test plan

- [ ] Verify `partialParse('{"key": "hello"}')` still works correctly
- [ ] Verify partial JSON with escaped quotes handles consecutive string tokens

### DIFF
--- a/src/_vendor/partial-json-parser/parser.ts
+++ b/src/_vendor/partial-json-parser/parser.ts
@@ -245,16 +245,22 @@ const tokenize = (input: string): Token[] => {
   },
   generate = (tokens: Token[]): string => {
     let output = '';
+    let prevType: string | null = null;
 
     tokens.map((token) => {
       switch (token.type) {
         case 'string':
-          output += '"' + token.value + '"';
+          if (prevType === 'string') {
+            output = output.slice(0, -1) + '\\"' + token.value + '"';
+          } else {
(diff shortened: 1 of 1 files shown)


localden:

After investigation, the bug this PR describes does not reproduce. The tokenizer at [`parser.ts:90-97`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/_vendor/partial-json-parser/parser.ts#L90-L97) correctly consumes `\"` into the token value, and the existing test at [`partial-json.test.ts:35`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/tests/lib/partial-json.test.ts#L35) (`partialParse('{"foo": "bar\\"baz"}')`) passes on `main`.

Consecutive string tokens only arise from already-invalid JSON like `{"a": "x" "y"}` — silently merging those would mask malformed model output rather than fix a parser bug.

If you have a reproduction case where valid JSON with escaped quotes fails, please open an issue with the exact input string. Closing for now.

PR #975, opened 2026-04-04
feat: add type guard helpers for ContentBlock variants

### LINKED ISSUE #432: [Types] Error: Property 'text' does not exist on type 'ContentBlock'.
After updating to `v0.22.0` from `v0.20.9`, I'm getting a type error on the last line of this block (`text = msg.content[0].text;`):

```ts
const msg = await anthropic.messages.create({
      model: "claude-3-opus-20240229",
      max_tokens: maxOutputTokens,
      temperature: 0.9,
      messages: [
        {
          role: "user",
          content: [
            {
              type: "text",
              text: prompt,
            },
          ],
        },
      ],
    });
    text = msg.content[0].text;
```

The type error is:
```
Property 'text' does not exist on type 'ContentBlock'.
  Property 'text' does not exist on type 'ToolUseBlock'.ts(2339)
```

When logging the message content, it does not appear to have changed, so I'm guessing this is just a type bug that needs fixed. My code works as expected when adding the `// @ts-ignore` comment above the problem line, which is my current workaround.

ac12644:
## Summary

- Adds 12 type guard functions (`isTextBlock`, `isToolUseBlock`, `isThinkingBlock`, etc.) that narrow `ContentBlock` to its specific variant
- Exported from the package root for ergonomic usage
- Includes compile-time exhaustiveness test that fails if a new `ContentBlock` variant is added without a corresponding guard

## Motivation

Addresses #432 — the most upvoted open issue (15+ thumbs up). Since v0.22.0, `ContentBlock` is a 12-member union, so `msg.content[0].text` requires type narrowing. Users currently resort to `as` casts or `// @ts-ignore`.

With this change:

```ts
import { isTextBlock } from '@anthropic-ai/sdk';

const message = await client.messages.create({ ... });
const textBlocks = message.content.filter(isTextBlock);
console.log(textBlocks[0].text); // fully typed, no casts
```

## Design decisions

- **Type guards over overloads**: Overloading `create()` based on `tools` presence would be type-unsound (thinking, server tools, etc. can produce non-text blocks even without `tools`). Type guards are correct and composable.
- **`src/lib/` placement**: Hand-written helpers that survive Stainless code generation, following the existing pattern (`parser.ts`, `MessageStream.ts`).
- **Exhaustiveness test**: Uses compile-time conditional types (same pattern as `TracksToolInput.test.ts`) so adding a new `ContentBlock` variant without a guard is a build error.

## Test plan

- [x] `yarn lint` passes (types + eslint + prettier)
- [x] 18 Jest tests pass (runtime behavior + type narrowing for all 12 guards)
- [x] Compile-time exhaustiveness check covers all 12 variants
- [x] Existing tests unaffected (`TracksToolInput`, `index.test`, `parser.test`)
- [x] Full build passes (CJS + ESM)
- [x] Verified imports work from built dist (`require` and `import`)

Closes #432

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,3 +29,18 @@ export type {
   ParseableMessageCreateParams,
   ExtractParsedContentFromParams,
 } from './lib/parser';
+
+export {
+  isTextBlock,
+  isThinkingBlock,
+  isRedactedThinkingBlock,
+  isToolUseBlock,
+  isServerToolUseBlock,
+  isWebSearchToolResultBlock,
+  isWebFetchToolResultBlock,
+  isCodeExecutionToolResultBlock,
+  isBashCodeExecutionToolResultBlock,
+  isTextEditorCodeExecutionToolResultBlock,
+  isToolSearchToolResultBlock,
+  isContainerUploadBlock,
+} from './lib/content-block-type-guards';
--- a/src/lib/content-block-type-guards.ts
+++ b/src/lib/content-block-type-guards.ts
@@ -1,0 +1,111 @@
+import type {
+  ContentBlock,
+  TextBlock,
+  ThinkingBlock,
+  RedactedThinkingBlock,
+  ToolUseBlock,
+  ServerToolUseBlock,
+  WebSearchToolResultBlock,
+  WebFetchToolResultBlock,
+  CodeExecutionToolResultBlock,
+  BashCodeExecutionToolResultBlock,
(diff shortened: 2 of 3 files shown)


dtmeadows-ant:
@ac12644 thanks, this is nicely done, exhaustiveness test included. We closed #432 as intended behavior, though. `content` is a discriminated union and TypeScript narrows it natively with `block.type === 'text'` (or an inline predicate in `.filter`).

PR #977, opened 2026-04-06
refactor: eliminate duplicate utility types, dead files, and shared parser runtime

apoc:
## What

Structural cleanup to eliminate duplicate code identified by static analysis (nucleus find_duplicate_code). All changes are behaviour-preserving — verified by 37/37 test suites, 584 tests.

### Commit 1 — Delete dead files with zero importers (`96a499b`)
- `src/internal/stream-utils.ts`: single export was a verbatim copy of `shims.ts`. Zero importers.
- `src/lib/tools/ToolRunner.ts`: orphaned predecessor of `BetaToolRunner.ts`, missing all features added since. Zero importers.

### Commit 2 — Consolidate duplicate utility types and functions (`cad969a`)
- **`src/lib/type-utils.ts`** (new): canonical home for `Simplify<T>` (4 file-private copies → 1) and `Promisable<T>` (re-exported from `BetaRunnableTool.ts` preserving existing import paths)
- **`NoInfer<T>`**: remove local workaround from both `json-schema.ts` files — TS 5.4+ ships it natively; project is on 5.8.3
- **`type Bytes`** in `core/streaming.ts`: import from `internal/decoders/line.ts` where it was already the canonical export
- **`isAbortError`**: re-export from `src/core/error.ts`; `packages/bedrock-sdk` replaces its byte-for-byte local copy with `import { isAbortError } from '@anthropic-ai/sdk/error'`

### Commit 3 — Extract shared parser runtime (`a38687f`)
- **`src/lib/core-parser.ts`** (new): `parseAccumulatedFormat(format, content)` — the 18-line parse step (json_schema check → `format.parse()` → `JSON.parse()` fallback → `AnthropicError` on failure) that was duplicated identically in `parser.ts` and `beta-parser.ts`. Both files now import and delegate to it.
- Type-level logic (conditional types, deprecated `output_format` shim, `.parsed` getter) stays in each file — intentionally different between stable and beta paths.

## Tests
37/37 suites, 584 passed, 1 skipped (pre-existing).


DepkaCZ (approved):

dtmeadows-ant:
@apoc thanks for the cleanup pass, but I don't think this is something we can accept as-is. Lots of the changes have / will conflict with future code-generation passes. I'll factor this into some refactors internally though to see what dead code we might could remove.

PR #980, opened 2026-04-08
fix: preserve directory paths in multipart uploads

daletyler1737:
## Problem

The SDK's multipart form handling strips directory paths from filenames when uploading files. The `getName` function in `internal/uploads.ts` uses `stripPath=true` by default in `createForm` and `multipartFormRequestOptions`.

This breaks `client.beta.skills.versions.create()` which requires `SKILL.md` to be inside a top-level directory (e.g. `my-skill/SKILL.md`).

## Solution

Change the default value of `stripFilenames` from `true` to `false` in both `createForm` and `multipartFormRequestOptions`.

## Related Issue

Fixes anthropics/anthropic-sdk-typescript#968

### DIFF
--- a/src/internal/uploads.ts
+++ b/src/internal/uploads.ts
@@ -83,7 +83,7 @@ type MultipartFormRequestOptions = Omit<RequestOptions, 'body'> & { body: unknow
 export const multipartFormRequestOptions = async (
   opts: MultipartFormRequestOptions,
   fetch: BaseAnthropic | Fetch,
-  stripFilenames: boolean = true,
+  stripFilenames: boolean = false,
 ): Promise<RequestOptions> => {
   return { ...opts, body: await createForm(opts.body, fetch, stripFilenames) };
 };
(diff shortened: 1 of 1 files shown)


localden:

This addresses issue [#968](https://github.com/anthropics/anthropic-sdk-typescript/issues/968), but flipping the global `stripFilenames` default in [`uploads.ts`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/internal/uploads.ts#L86) would also change behavior for [`files.upload()`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/resources/beta/files.ts#L137-L148), which would start sending full filesystem paths instead of basenames.

The targeted fix is to pass `false` at the `versions.create()` call site, mirroring what [`skills.create()` already does](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/resources/beta/skills/skills.ts#L54). Since that file is Stainless-generated, we're routing this through the codegen config — see [#1009](https://github.com/anthropics/anthropic-sdk-typescript/pull/1009) for tracking.

Closing in favor of the targeted approach.

PR #981, opened 2026-04-08
fix: propagate container.id across toolRunner iterations

daletyler1737:
## Summary

Fix two related bugs when using  with :

1. **Bug 1:**  is not automatically forwarded from one iteration to the next
2. **Bug 2:** The workaround using  either fails immediately or causes infinite loop

**Root Cause:** When using , each response includes a  that must be passed to subsequent requests. The current implementation doesn't extract and forward this .

**Fix:** After receiving a message response, extract the  (if present) and store it in  for use in subsequent message creations.

**Code Change:**
```typescript
// Forward container.id from response to next request for code_execution_20250120
if (this.#message?.container?.id) {
  this.#state.params.container = {
    id: this.#message.container.id,
    skills: (this.#message.container as any).skills ?? undefined,
  };
}
```

Fixes anthropics/anthropic-sdk-typescript#964

### DIFF
--- a/src/internal/uploads.ts
+++ b/src/internal/uploads.ts
@@ -83,7 +83,7 @@ type MultipartFormRequestOptions = Omit<RequestOptions, 'body'> & { body: unknow
 export const multipartFormRequestOptions = async (
   opts: MultipartFormRequestOptions,
   fetch: BaseAnthropic | Fetch,
-  stripFilenames: boolean = true,
+  stripFilenames: boolean = false,
 ): Promise<RequestOptions> => {
   return { ...opts, body: await createForm(opts.body, fetch, stripFilenames) };
 };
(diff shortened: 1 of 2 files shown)


localden:

This addresses the same bug as [#966](https://github.com/anthropics/anthropic-sdk-typescript/pull/966) and [#974](https://github.com/anthropics/anthropic-sdk-typescript/pull/974) (tracked in issue [#964](https://github.com/anthropics/anthropic-sdk-typescript/issues/964)). We're consolidating on [#966](https://github.com/anthropics/anthropic-sdk-typescript/pull/966), which includes test coverage and also fixes the related `setMessagesParams` over-invalidation.

This PR also bundles an unrelated `src/internal/uploads.ts` change — for that issue see the discussion on [#1009](https://github.com/anthropics/anthropic-sdk-typescript/pull/1009).

Closing as duplicate of [#966](https://github.com/anthropics/anthropic-sdk-typescript/pull/966).

PR #989, opened 2026-04-10
fix: handle invalid JSON escape sequences in partial-json-parser

### LINKED ISSUE #986: partialParse in _vendor/partial-json-parser crashes on invalid JSON escape sequences from model output
## Bug

When a model emits tool call JSON containing characters like unescaped backslashes (e.g. PHP namespaces `App\Http\Middleware\HandleAppearance`), the `partialParse` function in `_vendor/partial-json-parser/parser.{js,mjs}` throws:

```
Bad escaped character in JSON at position 80 (line 1 column 81)
```

This crashes the `MessageStream` during `input_json_delta` accumulation (line 501 in `MessageStream.mjs`), killing the entire streaming response.

## Root Cause

The `generate()` function reconstructs JSON from tokens but does not sanitize string values. The tokenizer correctly captures escape sequences like `\H` from `App\Http`, but `generate()` outputs them verbatim back into a JSON string:

```js
// generate() outputs: "use App\Http\Middleware" — \H and \M are invalid JSON escapes
output += ' + token.value + ';
```

`JSON.parse()` then rejects `\H`, `\M`, `\A` etc. since only `\" \\ \/ \b \f \n \r \t \uXXXX` are valid JSON escape sequences.

## Fix

In the `generate()` function, sanitize string token values by doubling backslashes that precede invalid escape characters:

```js
case 'string':
    // Repair invalid JSON escapes (e.g. \H from PHP namespaces)
    let sanitized = token.value.replace(/\\([^"\\\/bfnrtu])/g, \\$1);
    output += " + sanitized + ";
    break;
```

This turns `\H` → `\\H`, `\M` → `\\M` etc., producing valid JSON that `JSON.parse()` accepts.

Additionally, the `partialParse()` call in `MessageStream.{js,mjs}` should be wrapped in a try/catch so a parse failure during streaming does not crash the entire response.

## Reproduction

```js
import { partialParse } from "./_vendor/partial-json-parser/parser.mjs";

// Simulates model outputting PHP namespace backslashes in edit tool JSON
const input = '{"edits": [{"oldText": "use App\\Http\\Middleware\\HandleAppearance;", "newText": ""}]}';
partialParse(input); // throws: Bad escaped character in JSON
```

JosephDoUrden:
## Summary

Fix `partialParse` crashing on invalid JSON escape sequences (e.g. `\H`, `\M` from PHP namespaces like `App\Http\Middleware`) by repairing them in the `generate()` function before `JSON.parse()`.

Fixes #986

## Problem

When a model emits tool call JSON containing unescaped backslashes (e.g. PHP namespaces), `partialParse` throws `Bad escaped character in JSON`, crashing `MessageStream` during `input_json_delta` accumulation and killing the entire streaming response.

The tokeniser correctly captures escape sequences like `\H`, but `generate()` outputs them verbatim into a JSON string. `JSON.parse()` then rejects them since only `\" \\ \/ \b \f \n \r \t \uXXXX` are valid JSON escapes.

## Fix

In `generate()`, sanitise string token values by doubling backslashes before invalid escape characters:

```ts
token.value.replace(/\\([^"\\\/bfnrtu])/g, '\\\\$1')
```

This turns `\H` into `\\H`, `\M` into `\\M`, etc., producing valid JSON.

## Changes

- `src/_vendor/partial-json-parser/parser.ts`: Sanitise invalid escapes in `generate()`
- `tests/lib/partial-json.test.ts`: Added tests for invalid escape sequences and mixed valid/invalid escapes

## Test plan

- [x] All 19 partial-json tests pass (17 existing + 2 new)
- [x] New test: PHP namespace backslashes (`App\Http\Middleware`) parse correctly
- [x] New test: Mixed valid (`\n`) and invalid (`\H`) escapes in same string

### DIFF
--- a/src/_vendor/partial-json-parser/parser.ts
+++ b/src/_vendor/partial-json-parser/parser.ts
@@ -249,7 +249,9 @@ const tokenize = (input: string): Token[] => {
     tokens.map((token) => {
       switch (token.type) {
         case 'string':
-          output += '"' + token.value + '"';
+          // Repair invalid JSON escape sequences (e.g. \H from PHP namespaces)
+          // by doubling backslashes before characters that are not valid JSON escapes
+          output += '"' + token.value.replace(/\\([^"\\\/bfnrtu])/g, '\\\\$1') + '"';
           break;
         default:
           output += token.value;
--- a/tests/lib/partial-json.test.ts
+++ b/tests/lib/partial-json.test.ts
@@ -78,4 +78,16 @@ describe('partialParse', () => {
   test('deeply nested partial JSON objects', () => {
     expect(partialParse(`{"a": {"b": {"c": {"d": "e`)).toEqual({ a: { b: { c: {} } } });
   });
+
+  test('string with invalid JSON escape sequences (e.g. PHP namespaces)', () => {
+    expect(partialParse(`{"code": "use App\\Http\\Middleware"}`)).toEqual({
+      code: 'use App\\Http\\Middleware',
+    });
+  });
+
+  test('string with mixed valid and invalid escape sequences', () => {
+    expect(partialParse(`{"text": "line1\\nApp\\Http\\nline2"}`)).toEqual({
+      text: 'line1\nApp\\Http\nline2',
+    });
+  });
 });


WGlynn:
Reproduced this locally against the PR. #989 fixes the reported case: `use App\Http\Middleware` parses cleanly instead of throwing `Bad escaped character in JSON`. Confirmed against the two tests in the PR as well.

There is a residual class it does not cover, though, and I think it points back to the second fix the issue asked for.

The sanitize regex only doubles a backslash when the following character is outside the valid-escape set (`" \ / b f n r t u`). That leaves a literal backslash before one of the valid-escape letters untouched, which is still wrong when the backslash was meant literally. Two real cases a model hits:

- `\u` not followed by four hex digits. A Windows-style path like `{"p":"C:\users\local"}` still throws `Bad Unicode escape` after this PR, because `\u` is treated as a valid escape and skipped. Same for a truncated `\u12` mid-stream.
- `\t`, `\n`, etc. before path text. `{"p":"C:\test"}` no longer crashes but silently parses to `C:` + tab + `est`, so the value is corrupted rather than repaired.

Repro (built from `src/_vendor/partial-json-parser/parser.ts` at this PR vs main):

```
input: {"p":"\users\local"}
  main  : throws  Bad Unicode escape in JSON at position 8
  #989  : throws  Bad Unicode escape in JSON at position 8   <- unchanged

input: {"p":"C:\Users\test"}
  main  : throws  Bad escaped character
  #989  : ok, but value is  C:\Users<TAB>est
```

The takeaway is not that the regex should grow to catch `\u`, since you can't enumerate every malformed escape a streamed delta might contain. It is that the issue's other suggestion, wrapping the `partialParse()` call in `MessageStream` in a try/catch, is the part that actually makes streaming robust: a single bad `input_json_delta` should degrade that one accumulation, not tear down the whole response. #989 is worth keeping because it removes the most common trigger (backslash before a non-escape char, the PHP-namespace case). Pairing it with the try/catch backstop closes the class rather than one instance of it.

Happy to send the minimal repro script if useful.

PR #936, opened 2026-03-08
perf(messagestream): optimize parsing performance with lazy getter  (Issue #933)

Emeenent14:
Converted `content.input` in `MessageStream` to a lazy getter. Previously, the SDK re-parsed the entire JSON buffer on every delta, leading to O(N²) complexity. This change ensures parsing only happens when needed, drastically improving responsiveness for large streaming responses.
 
 Closes #933

PR #997, opened 2026-04-12
perf: fix O(N^2) accumulation in MessageStream and BetaMessageStream

### LINKED ISSUE #933: O(N²) performance bottleneck in [MessageStream](src/lib/MessageStream.ts#51-747) during structured output streaming

### LINKED ISSUE #995: perf(MessageStream): O(N²) complexity in text accumulation during structured output streaming
### Description
`MessageStream` accumulates streamed text using string concatenation (`accumulated += chunk`), which is O(N²) in JavaScript due to string immutability. Each concatenation copies the entire accumulated string. For long responses (code generation, structured JSON output), this causes severe performance degradation.

### Reproduction

```typescript
import { performance } from 'node:perf_hooks';

async function benchmarkStreaming(targetTokens: number) {
  const start = performance.now();

    model: 'claude-opus-4-6',
    max_tokens: targetTokens,
    messages: [{
      role: 'user',
      content: `Generate exactly ${targetTokens} tokens of valid JSON. Start now:`
    }],
  });

  for await (const chunk of stream) {
    // just consume the stream
  }

  const elapsed = performance.now() - start;
  const finalMessage = await stream.finalMessage();
  const actualTokens = finalMessage.usage.output_tokens;

  console.log(`${actualTokens} tokens: ${elapsed.toFixed(0)}ms`);
}

// Compare: 1K vs 4K tokens — should be 4x, actually ~16x
await benchmarkStreaming(1000);
await benchmarkStreaming(4000);
```

### Expected Behavior
Throughput should scale **linearly** with response size. Streaming 4K tokens should take approximately 4x longer than 1K tokens.

### Actual Behavior
Due to O(N²) complexity, 4K tokens takes approximately **16x** longer than 1K tokens. At 10K tokens, performance is ~100x slower than optimal.

### Root Cause
In `src/lib/MessageStream.ts`, the `accumulateContent` method (or equivalent) does:
```typescript
// O(N²) — string concatenation in a loop
snapshot.content[i].text = (snapshot.content[i].text ?? '') + delta.text;
```

Each `+=` on a string creates a NEW string by copying all existing characters plus the new characters. After N chunks of average size C, total work done = C + 2C + 3C + ... + NC = O(N²·C).

### Proposed Fix
Use an array of chunks and join once when the stream completes:
```typescript
// Internal: accumulate into an array
const textChunks: string[] = [];
textChunks.push(delta.text);

// When stream ends: materialize once
const finalText = textChunks.join('');
```

### Environment
- `@anthropic-ai/sdk` version: latest
- Node.js version: 20.x / 22.x
- Affected models: All (especially with structured outputs / long generations)

### Impact
- **Severity:** High — affects all users generating responses > 2K tokens
- **Workaround:** None (the accumulation is internal to the SDK)
- **Regression risk:** Low — pure performance optimization with no behavior change

I'm happy to submit a PR with the fix + benchmarks. 🙏

claygeo:
## Summary

Replaces object spread with direct in-place mutation in `#accumulateMessage` for all `content_block_delta` types in both `MessageStream` and `BetaMessageStream`. The previous implementation created a new content block object via `{ ...snapshotContent, prop: newValue }` on every delta event, causing O(N^2) work for N deltas.

**Delta types fixed:**
- `text_delta`: `{ ...obj, text: old + new }` → `obj.text = old + new`
- `citations_delta`: `[...existing, new]` → `Array.push()` (O(N^2) → O(N) for array copy)
- `input_json_delta`: spread + `defineProperty` each time → `defineProperty` once, then direct assignment
- `thinking_delta`: `{ ...obj, thinking: old + new }` → `obj.thinking = old + new`
- `signature_delta`: `{ ...obj, signature: val }` → `obj.signature = val`
- `compaction_delta` (Beta only): `{ ...obj, content: old + new }` → `obj.content = old + new`

**Why this is safe:** The `message_delta` case already uses direct mutation (`snapshot.stop_reason = event.delta.stop_reason`). All content block TypeScript types are mutable interfaces (no `readonly` modifiers). The snapshot is stored in a private field and events emit references to the same object.

**Also fixes:** BetaMessageStream's `content_block_start` was missing the defensive copy (`{ ...event.content_block }`) that MessageStream has. Added for consistency since we now mutate content blocks in place.

Fixes #995
Fixes #933

## Relationship to PR #936

PR #936 addresses only `input_json_delta` with a lazy getter approach. This PR fixes all 6 delta types in both files using a simpler, consistent mutation pattern.

## Test plan

- [x] All existing tests pass (`MessageStream.test.ts`, `BetaMessageStream.test.ts`, `streaming.test.ts`)
- [x] Added test: "accumulates large text responses in linear time" — streams 5,000 text deltas and verifies correct accumulation
- [x] `tsc --noEmit` passes with no type errors

### DIFF
--- a/src/lib/BetaMessageStream.ts
+++ b/src/lib/BetaMessageStream.ts
@@ -611,40 +611,39 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
         switch (event.delta.type) {
           case 'text_delta': {
             if (snapshotContent?.type === 'text') {
-              snapshot.content[event.index] = {
-                ...snapshotContent,
-                text: (snapshotContent.text || '') + event.delta.text,
-              };
+              snapshotContent.text = (snapshotContent.text || '') + event.delta.text;
             }
             break;
           }
           case 'citations_delta': {
             if (snapshotContent?.type === 'text') {
-              snapshot.content[event.index] = {
-                ...snapshotContent,
-                citations: [...(snapshotContent.citations ?? []), event.delta.citation],
-              };
+              if (!snapshotContent.citations) {
+                snapshotContent.citations = [];
+              }
+              snapshotContent.citations.push(event.delta.citation);
             }
             break;
           }
           case 'input_json_delta': {
             if (snapshotContent && tracksToolInput(snapshotContent)) {
-              // we need to keep track of the raw JSON string as well so that we can
-              // re-parse it for each delta, for now we just store it as an untyped
-              // non-enumerable property on the snapshot
+              // Track the raw JSON string as a non-enumerable property so it's
+              // hidden from JSON.stringify/serialization but available for re-parsing
               let jsonBuf = (snapshotContent as any)[JSON_BUF_PROPERTY] || '';
               jsonBuf += event.delta.partial_json;
 
-              const newContent = { ...snapshotContent };
-              Object.defineProperty(newContent, JSON_BUF_PROPERTY, {
-                value: jsonBuf,
-                enumerable: false,
-                writable: true,
-              });
+              if (!(JSON_BUF_PROPERTY in snapshotContent)) {
+                Object.defineProperty(snapshotContent, JSON_BUF_PROPERTY, {
+                  value: jsonBuf,
+                  enumerable: false,
+                  writable: true,
+                });
+              } else {
+                (snapshotContent as any)[JSON_BUF_PROPERTY] = jsonBuf;
+              }
 
               if (jsonBuf) {
                 try {
-                  newContent.input = partialParse(jsonBuf);
+                  (snapshotContent as any).input = partialParse(jsonBuf);
                 } catch (err) {
                   const error = new AnthropicError(
                     `Unable to parse tool parameter JSON from model. Please retry your request or adjust your prompt. Error: ${err}. JSON: ${jsonBuf}`,
@@ -652,34 +651,24 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
                   this.#handleError(error);
                 }
               }
-              snapshot.content[event.index] = newContent;
             }
             break;
           }
           case 'thinking_delta': {
             if (snapshotContent?.type === 'thinking') {
-              snapshot.content[event.index] = {
-                ...snapshotContent,
-                thinking: snapshotContent.thinking + event.delta.thinking,
-              };
+              snapshotContent.thinking = snapshotContent.thinking + event.delta.thinking;
             }
             break;
           }
           case 'signature_delta': {
             if (snapshotContent?.type === 'thinking') {
-              snapshot.content[event.index] = {
-                ...snapshotContent,
-                signature: event.delta.signature,
-              };
+              snapshotContent.signature = event.delta.signature;
             }
             break;
           }
           case 'compaction_delta': {
             if (snapshotContent?.type === 'compaction') {
-              snapshot.content[event.index] = {
-                ...snapshotContent,
-                content: (snapshotContent.content || '') + event.delta.content,
-              };
+              snapshotContent.content = (snapshotContent.content || '') + event.delta.content;
             }
             break;
           }
--- a/src/lib/BetaMessageStream.ts
+++ b/src/lib/BetaMessageStream.ts
@@ -603,7 +603,7 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
 
         return snapshot;
       case 'content_block_start':
-        snapshot.content.push(event.content_block);
+        snapshot.content.push({ ...event.content_block });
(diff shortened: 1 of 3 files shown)


minglong51:
Hi @claygeo, @localden — thanks for the cleanup here, the object-spread removal is a real allocation win (and the fact that it eliminates the redundant `snapshot.content[event.index] = newContent` for `input_json_delta` is nice).

I wanted to flag a remaining piece for the maintainer's call: the string concatenation itself (`(snapshotContent.text || '') + event.delta.text` on lines 614 / 616 etc) is still the underlying O(N²) operation the issue body refers to. V8's cons-string optimization papers over it for small-to-medium streams, but it breaks down at larger N. Synthetic bench on Node 22.14.0, median of 3 runs, 5-char chunks (close to a real token delta), per-event char access to simulate a progressive-UI consumer:

| N text_deltas | post-#997 `+=` (ms) | array-buffer + finalText (ms) |
|---|---|---|
| 1,000 | 0.5 | 0.0 |
| 4,000 | 2.0 | 0.1 |
| 10,000 | 10.0 | 0.2 |
| 20,000 | 24.5 | 0.4 |
| 40,000 | 282.1 | 0.7 |

At 40k deltas the `+=` pattern is ~400× slower than an array-buffer that joins once at finalization. The 20k→40k jump is 11.5× for 2× input — V8 transitioning out of cons-string into quadratic territory. 40k events is in range for long structured-output generations (a 16k-token JSON tool output emitted at ~3 chars/delta lands there).

The complementary fix would be a chunks-array on a Symbol-keyed non-enumerable property + a lazy enumerable getter on `.text` (and `.thinking`, `.content`) that joins on read. Mirrors the existing `JSON_BUF_PROPERTY` pattern in this same method. Helper is ~25 lines; touches the same 5 sites this PR already touches.

One honest caveat I'd want to surface: this only helps consumers that read `.text` infrequently (typically via `finalText()` / `finalMessage()`). For consumers listening to the per-delta `'text'` event — `_emit('text', event.delta.text, content.text || '')` at `MessageStream.ts:462` and the Beta equivalent — the second arg forces a flatten/materialize on every event, and neither `+=` nor array-buffer can fully fix that without changing the event's second-arg signature (likely breaking).

Happy to draft this as a follow-up PR (or fold into this one if you'd prefer to land both at once). Either way — wanted to put the numbers in front of you before this lands so it's an informed call.

<details>
<summary>Bench script (Node, no deps)</summary>

```js
import { performance } from 'node:perf_hooks';
function benchConcat(n, k) {
  let acc = ''; const chunk = 'a'.repeat(k); let sink = 0;
  const t0 = performance.now();
  for (let i = 0; i < n; i++) {
    acc = (acc || '') + chunk;
    sink += acc.charCodeAt(acc.length - 1);
  }
  return performance.now() - t0;
}
function benchArray(n, k) {
  const chunks = []; const chunk = 'a'.repeat(k);
  const t0 = performance.now();
  for (let i = 0; i < n; i++) chunks.push(chunk);
  chunks.join('');
  return performance.now() - t0;
}
for (const n of [1000, 4000, 10000, 20000, 40000]) {
  console.log(n, benchConcat(n, 5).toFixed(1), benchArray(n, 5).toFixed(1));
}
```
</details>

claygeo:
Closing this in favor of #1096.

I rebased onto current `main` and re-checked whether this change still helps. It doesn't hold up, for three reasons:

1. **#933 is already fixed on `main`.** Commit `65a0106` added `withLazyInput()`, a lazy memoized getter for `.input`, so `partialParse` now runs once per block instead of once per delta. This PR's `input_json_delta` change would revert that.

2. **The accumulate change targets the wrong layer.** I benchmarked it: object-spread vs in-place mutation in `#accumulateMessage` is roughly 0.1% of streaming cost. The real O(N²) is in `iterSSEChunks` (`src/core/streaming.ts`), where `data = data.slice(patternIndex)` re-copies the whole remaining buffer once per event. A consumer that never reads a snapshot still goes quadratic, which rules out `#accumulateMessage` as the cause.

3. **In-place mutation reverts `ae63aa6`** ("prevent mutations in message streaming"), which was a deliberate design choice.

The real fix is #1096: an offset cursor in `iterSSEChunks`, byte-identical output, 16k events 3.5s → 0.2s, with regression tests. Thanks for your patience with the misdiagnosis here.

PR #1006, opened 2026-04-17
fix(stream): handle invalid JSON escape sequences in MessageStream

### LINKED ISSUE #996: bug(streaming): JSON.parse crashes on invalid escape sequences in model output
### Description
When Claude outputs a string containing certain escape sequences that are valid in some contexts but not strict JSON (e.g., `\d`, `\p`, `\s` as regex literals inside a JSON string), the streaming parser crashes with an unhandled `SyntaxError`. This surfaces as an unhandled rejection that crashes the entire application.

### Reproduction

```typescript

// Ask Claude to produce output containing regex-like content
  model: 'claude-sonnet-4-6',
  max_tokens: 500,
  messages: [{
    role: 'user',
    content: 'Return a JSON object with a "pattern" field containing the regex \\d+\\.\\d+ as a string value. Return ONLY the JSON.'
  }],
});

// This may crash: SyntaxError: Invalid escape sequence
const message = await stream.finalMessage();
```

### Expected Behavior
The SDK should gracefully handle malformed JSON from the model, either by:
1. Attempting lenient JSON parsing (JSON5 / repair strategies)
2. Returning the raw string with a warning
3. Emitting a recoverable error event instead of throwing

### Actual Behavior
SyntaxError: Invalid or unexpected token
at JSON.parse (<anonymous>)
at MessageStream.parse (src/lib/MessageStream.ts:XXX)
...

The error is unhandled and crashes the Node.js process.

### Proposed Fix
Wrap all `JSON.parse` calls in the streaming pipeline with a recovery strategy:
```typescript
function safeParseJSON<T>(text: string): T | null {
  try {
    return JSON.parse(text) as T;
  } catch {
    // Attempt to repair common model output issues
    const repaired = repairInvalidEscapes(text);
    try {
      return JSON.parse(repaired) as T;
    } catch {
      return null; // Emit warning, don't crash
    }
  }
}
```

### Environment
- `@anthropic-ai/sdk` version: latest
- Node.js version: 20.x
- Trigger: Models asked to produce JSON containing regex patterns, Windows paths, LaTeX

I'm happy to submit a PR. 🙏

xodn348:
## Summary

Fixes #996

When the Anthropic API streams tool-use deltas containing invalid JSON escape sequences (e.g. `\d`, `\p`, `\s` from regex patterns), `partialParse()` throws an unhandled error that crashes the entire `MessageStream`. This is a runtime crash with no way for consumers to recover gracefully.

## Change

Wrapped the `partialParse(jsonBuf)` call in `#accumulateMessage` with a try-catch that silently skips the delta on parse failure. The stream continues processing subsequent events, and the final complete JSON (which uses proper `\\d` escaping) parses correctly.

**Before:**
```typescript
newContent.input = partialParse(jsonBuf);
```

**After:**
```typescript
try {
  newContent.input = partialParse(jsonBuf);
} catch {
  // invalid escape sequences (e.g. \d, \p, \s) — skip this delta, stream continues
}
```

This is the minimal surgical fix — one file changed, 5 lines added, 1 removed.

## Verification

- `tsc --noEmit` passes (EXIT_CODE=0)
- `npm test` passes (641 passed, 5 failed pre-existing Nock connection errors in ToolRunnerE2E unrelated to this change)

### DIFF
--- a/src/lib/MessageStream.ts
+++ b/src/lib/MessageStream.ts
@@ -637,7 +637,11 @@ export class MessageStream<ParsedT = null> implements AsyncIterable<MessageStrea
               });
 
               if (jsonBuf) {
-                newContent.input = partialParse(jsonBuf);
+                try {
+                  newContent.input = partialParse(jsonBuf);
+                } catch {
+                  // invalid escape sequences (e.g. \d, \p, \s) — skip this delta, stream continues
+                }
               }
               snapshot.content[event.index] = newContent;
             }


localden:

We're consolidating on [#989](https://github.com/anthropics/anthropic-sdk-typescript/pull/989), which fixes the root cause in the partial-json parser itself ([`_vendor/partial-json-parser/parser.ts`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/_vendor/partial-json-parser/parser.ts)) rather than catching at the call site.

Separately, if call-site protection is desired, it should mirror the existing pattern in [`BetaMessageStream.ts:646-653`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/lib/BetaMessageStream.ts#L646-L653) (`AnthropicError` + `#handleError`) rather than a silent swallow.

Closing as duplicate of [#989](https://github.com/anthropics/anthropic-sdk-typescript/pull/989).

PR #1015, opened 2026-04-26
fix(MessageStream): propagate container from message_delta into final snapshot

abhicris:
## What

`MessageStream.#accumulateMessage()` (`src/lib/MessageStream.ts`) merges `stop_reason`, `stop_sequence`, and `usage` from the `message_delta` SSE event into the running snapshot but never copies `event.delta.container`. With `code_execution_20260120` (or any container-backed tool), `stream.finalMessage().container` is therefore always `null`, even though the SSE stream carried the container info — and the next request to the same conversation fails with:

> 400 — `container_id is required when there are pending tool uses generated by code execution with tools.`

## Why

The beta path (`src/lib/BetaMessageStream.ts:578`) already does the right thing:

```ts
case 'message_delta':
  snapshot.container = event.delta.container;
  snapshot.stop_reason = event.delta.stop_reason;
  ...
```

This PR mirrors that single line in the non-beta path. Both `Message.container` (`messages.ts:887`) and `RawMessageDeltaEvent.Delta.container` (`messages.ts:1217`) are already typed as `Container | null`, so this is a pure assignment with no API or type-shape change.

## Refs

- #964 — surfaced in @pierrevalade's [comment](https://github.com/anthropics/anthropic-sdk-typescript/issues/964#issuecomment-) as a separate, lower-level bug from the four open `BetaToolRunner` PRs (#966, #974, #981, #994), affecting anyone using `messages.stream()` + `finalMessage()` directly. Those PRs handle the runner-level propagation; this PR fixes the underlying snapshot accumulation that those (and any other consumer) implicitly rely on.
- Mirrors anthropics/anthropic-sdk-python#1444 (same fix in the Python SDK).

## Test plan

- New fixture `tests/lib/fixtures/code_execution_response.txt` emits a `message_delta` carrying a `container` block.
- New test `propagates container from message_delta into final snapshot` drains the stream and asserts `finalMessage.container.id` and `expires_at` match.
- The test correctly **fails on `main`** with `Expected: {id: ..., expires_at: ...}, Received: null`, and **passes with this fix**.
- Full `tests/api-resources/MessageStream.test.ts` suite still passes (7/7).

## Diff size

`+50 / 0` (1-line fix, 31-line test, 18-line fixture).


tomer-ant:
Thanks, but this will be fixed in the next release: https://github.com/anthropics/anthropic-sdk-typescript/pull/1146/changes/7b82659d70201ae3edab846f1693366db7e62ce3

PR #1040, opened 2026-05-12
fix(helpers): give a clear error when a zod v3 schema is passed to the zod helpers

### LINKED ISSUE #1035: zodOutputFormat fails when project uses zod3
Stackblitz: https://stackblitz.com/edit/vitejs-vite-ntcvltqj?file=src%2Fmain.ts&view=editor

src/main.ts
```typescript
import { zodOutputFormat } from '@anthropic-ai/sdk/helpers/zod';
import zod3 from 'zod';
import zod4 from 'zod/v4';

const Schema3 = zod3.object({
  foo: zod3.string(),
});

const Schema4 = zod4.object({
  foo: zod4.string(),
});

try {
  zodOutputFormat(Schema3);
} catch {
  console.log('zod 3 schema fails');
}

// @ts-expect-error typings are broken somehow
zodOutputFormat(Schema4);
console.log('zod 4 schema works');
```

```
~/projects/vitejs-vite-ntcvltqj
❯ node src/main.ts
zod 3 schema fails
zod 4 schema works
```

Zawwarsami16:
The zod helpers all call `z.toJSONSchema` imported from `zod/v4`, but the package's peerDeps say `zod ^3.25.0 || ^4.0.0` is supported. When someone running against zod 3.x does `import { z } from 'zod'` and passes that schema into `zodOutputFormat`/`betaZodOutputFormat`/`betaZodTool`, the call dies inside the zod v4 internals with a confusing `TypeError: Cannot read properties of undefined (reading 'def')`. Repro from #1035 is a few lines:

```ts
import { zodOutputFormat } from '@anthropic-ai/sdk/helpers/zod';
import zod3 from 'zod';
zodOutputFormat(zod3.object({ foo: zod3.string() })); // throws TypeError
```

I traced this to the shape difference: zod v4 schemas always have a `_zod` property (the new internal namespace), zod v3 schemas don't. `z.toJSONSchema` reaches into `_zod.def` immediately, which is where the `undefined`-deref comes from. The peerDeps wording is honest in spirit because zod 3.25+ ships a `zod/v4` subpath as a backport, so a v3-installed project can import `zod/v4` and get a v4-shaped schema. The user just has to know that's the contract.

So this PR doesn't try to make v3 schemas work (that'd want a different JSON Schema converter and a new dep). It just adds a `src/lib/check-zod-version.ts` helper that detects the v3 shape and throws an `AnthropicError` pointing at the two unblock paths. Wired into the three callsites in `src/helpers/zod.ts` and `src/helpers/beta/zod.ts`. Tests cover all three functions with a v3-shaped mock so we can exercise the detection without juggling two zod installs in dev.

Closes #1035

Test plan:
- [x] `yarn jest tests/helpers/zod.test.ts` passes (23/23, 20 existing + 3 new v3-rejection cases)
- [x] Confirmed against a real `zod@3.25.76` schema in a scratch project that the new error message fires instead of the `TypeError`
- [x] `yarn lint` clean on the touched files

### DIFF
--- a/src/helpers/beta/zod.ts
+++ b/src/helpers/beta/zod.ts
@@ -1,6 +1,7 @@
 import { transformJSONSchema } from '../..//lib/transform-json-schema';
 import * as z from 'zod/v4';
 import { AnthropicError } from '../../core/error';
+import { assertZodV4Schema } from '../../lib/check-zod-version';
 import { AutoParseableBetaOutputFormat } from '../../lib/beta-parser';
 import { BetaRunnableTool, BetaToolRunContext, Promisable } from '../../lib/tools/BetaRunnableTool';
 import { BetaToolResultContentBlockParam } from '../../resources/beta';
@@ -16,6 +17,7 @@ import { BetaToolResultContentBlockParam } from '../../resources/beta';
 export function betaZodOutputFormat<ZodInput extends z.ZodType>(
   zodObject: ZodInput,
 ): AutoParseableBetaOutputFormat<z.infer<ZodInput>> {
+  assertZodV4Schema(zodObject, 'betaZodOutputFormat');
   let jsonSchema = z.toJSONSchema(zodObject, { reused: 'ref' });
 
   jsonSchema = transformJSONSchema(jsonSchema);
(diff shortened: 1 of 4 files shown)


Zawwarsami16:
rebased this onto current main to clear the conflict. checked while doing it and the underlying issue is still there, src/helpers/zod.ts and the beta one still call z.toJSONSchema straight from zod/v4 with no guard, so passing a v3 schema still blows up with the cryptic "Cannot read properties of undefined (reading 'def')". the assert just turns that into an actionable message. tests pass locally.

PR #1049, opened 2026-05-17
fix(stream): detect mixed newline SSE boundaries

pragnyanramtha:
## Summary

- detect SSE chunk boundaries made from mixed adjacent line endings such as `\n\r\n` and `\r\n\n`
- replace the hard-coded double-newline byte patterns with a small line-ending scanner
- add focused regression coverage for mixed adjacent line endings

## Context

The SSE chunk scanner previously recognized only `\n\n`, `\r\r`, and `\r\n\r\n`. Event streams allow lines to end with `\r`, `\n`, or `\r\n`, so mixed adjacent line endings are valid blank-line boundaries. Without detecting those boundaries, a live stream using a mixed separator can keep events buffered until another recognized separator or the response close.

## Validation

- `./node_modules/.bin/jest tests/internal/decoders/line.test.ts --runInBand`
- `./node_modules/.bin/jest tests/streaming.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/internal/decoders/line.ts tests/internal/decoders/line.test.ts`
- `./node_modules/.bin/eslint src/internal/decoders/line.ts tests/internal/decoders/line.test.ts`
- `./node_modules/typescript/bin/tsc --noEmit`
- `git diff --check`

Note: `yarn install --frozen-lockfile` failed because the current lockfile needs dependency metadata updates; a plain `yarn install` was used for local validation, and install-only lockfile churn was excluded from this PR.

### DIFF
--- a/src/internal/decoders/line.ts
+++ b/src/internal/decoders/line.ts
@@ -104,31 +104,34 @@ function findNewlineIndex(
 }
 
 export function findDoubleNewlineIndex(buffer: Uint8Array): number {
-  // This function searches the buffer for the end patterns (\r\r, \n\n, \r\n\r\n)
-  // and returns the index right after the first occurrence of any pattern,
-  // or -1 if none of the patterns are found.
+  // This function searches the buffer for two adjacent line endings and returns
+  // the index right after the second one, or -1 if none are found. SSE allows
+  // lines to end in \r, \n, or \r\n, so adjacent line endings may be mixed.
+  for (let i = 0; i < buffer.length; i++) {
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@pragnyanramtha thanks. Our API terminates SSE events with a consistent separator, and a boundary made of mixed adjacent endings (`\n\r\n` and the like) doesn't occur on a first-party stream. Where we've seen odd framing it's been an intermediary rewriting it. I'm going to close rather than loosen the scanner speculatively. If you've hit buffered events against the API directly, a `request-id` would make the case.

PR #1073, opened 2026-05-28
fix: propagate container.id across toolRunner iterations

### LINKED ISSUE #964: Bug: `toolRunner`'s `setMessagesParams` causes duplicate tool call loops

Sarmkadan:
## Summary

Fixes #964 — `toolRunner` does not propagate `container.id` across iterations, causing a hard error when using `code_execution` with client-side `betaTool`s.

### Root cause

When the API returns a response with a `container` object (created by `code_execution`), the `container.id` must be included in subsequent requests so the API can reuse the same container. Neither `BetaToolRunner` nor the legacy `ToolRunner` extracted this value from the response, so the next iteration sent a request without `container`, triggering:

```json
{"type":"invalid_request_error","message":"container_id is required when there are pending tool uses generated by code execution with tools."}
```

The only workaround (`setMessagesParams`) also didn't work correctly — setting `container` via the mutator triggered the `#mutated` flag, which prevented the assistant message from being pushed to the conversation history, causing an infinite loop (Bug 2 in the issue).

### Fix

After awaiting each response, extract `container.id` and set it on `this.#state.params.container` before the `#mutated` check. This ensures:

1. The container is propagated regardless of whether the user called `setMessagesParams`
2. Users no longer need the broken `setMessagesParams` workaround
3. The fix is minimal — 4 lines of logic in each runner

Applied to both `BetaToolRunner` and the legacy `ToolRunner`.

## Test plan

- [x] `forwards container.id from response to next request` — non-streaming mode, verifies second request body contains `container`
- [x] `forwards container.id in streaming mode` — same verification via SSE streaming
- [x] All 30 existing ToolRunner tests pass

### DIFF
--- a/src/lib/tools/BetaToolRunner.ts
+++ b/src/lib/tools/BetaToolRunner.ts
@@ -196,11 +196,15 @@ export class BetaToolRunner<Stream extends boolean> {
             yield this.#message as any;
           }
 
+          const message = await this.#message;
+          if (message.container?.id) {
+            this.#state.params.container = message.container.id;
+          }
+
           const isCompacted = await this.#checkAndCompact();
           if (!isCompacted) {
             if (!this.#mutated) {
-              const { role, content } = await this.#message;
-              this.#state.params.messages.push({ role, content });
+              this.#state.params.messages.push({ role: message.role, content: message.content });
             }
 
             const toolMessage = await this.#generateToolResponse(this.#state.params.messages.at(-1)!);
--- a/src/lib/tools/ToolRunner.ts
+++ b/src/lib/tools/ToolRunner.ts
@@ -107,9 +107,13 @@ export class BetaToolRunner<Stream extends boolean> {
             yield this.#message as any;
           }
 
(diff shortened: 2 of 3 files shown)


tomer-ant:
Thanks, but this is fixed in https://github.com/anthropics/anthropic-sdk-typescript/issues/1114#issuecomment-5415100070

PR #1079, opened 2026-05-31
fix: update 'anthropic-beta' header condition to check for betas length

suekou:
## Summary

Passing `betas: []` sends an empty `anthropic-beta:` header instead of omitting it, and the API rejects the request with a 400:

> Unexpected value(s) `` for the `anthropic-beta` header. Please consult our documentation at docs.claude.com or try again without the header.

`betas: undefined` already omits the header. An empty array means "no betas", so it should behave the same way.

## Cause

The header is built with `betas?.toString() != null`. For `[]`, `[].toString()` is `""`, which passes the `!= null` check and sets the header to an empty string. (`undefined?.toString()` is `undefined`, so that path is correctly skipped.)

## Fix

Guard on `betas?.length` instead, so both `undefined` and `[]` skip the header and only a non-empty array sets it. Applied to the endpoints that don't append a mandatory beta of their own:

- `completions.create`
- `models.retrieve` / `models.list`
- `beta.models.retrieve` / `beta.models.list`
- `beta.messages.create`

Endpoints that always append a required beta (files, batches, etc.) are unaffected — they never produce an empty header.


dtmeadows-ant:
@suekou thanks, confirmed: `betas: []` produces an empty `anthropic-beta` header and the API 400s on it. That `betas?.toString() != null` guard is emitted for every array-valued header param across the generated resource methods. The fix belongs in the generator (skip empty arrays) so all of them get it, rather than patching four files by hand; noted with your analysis. Until then, pass `undefined` instead of `[]`.

PR #1100, opened 2026-06-24
fix: honor request timeout for long non-streaming messages

### LINKED ISSUE #1097: Long Request error is thrown even when timeout is specified in request options (but not on the client)
If a request has a large max budget and the [`calculateNonstreamingTimeout() function`](https://github.com/anthropics/anthropic-sdk-typescript/blob/ab700dc013e735b80dead115fcae52fba83f1aeb/src/client.ts#L1390) estimates that it might take more than 10 minutes, it raises an error even when the request timeout (but not client timeout) has been specified.

The exact message is:

> Streaming is required for operations that may take longer than 10 minutes. See https://github.com/anthropics/anthropic-sdk-typescript#long-requests for more details

The link is out of date, but the [moved documentation](https://platform.claude.com/docs/en/cli-sdks-libraries/sdks/typescript#long-requests) explicitly says:

> Passing `stream: true` or overriding the `timeout` option at the client or request level disables this error.

The culprit appears to be located in the [`Request.create() function`](https://github.com/anthropics/anthropic-sdk-typescript/blob/ab700dc013e735b80dead115fcae52fba83f1aeb/src/resources/messages/messages.ts#L88), where it accesses `this._client._options.timeout` but ignores the `options` parameter. Simply appending `?? options?.timeout` to the end of the expression would hopefully solve it.

lin-hongkuan:
## Summary
Honor request-level `timeout` options before calculating long non-streaming message timeouts.

## Motivation
Passing a client-level timeout suppresses the long non-streaming request timeout calculation, but passing the same timeout per request currently does not. This causes a long-request error even when callers explicitly provide a request-level timeout.

Fixes #1097

## Changes
- Prefer `options.timeout` before falling back to the client-level timeout in stable messages.
- Apply the same behavior to beta messages.
- Add stable and beta tests asserting request-level timeout bypasses `calculateNonstreamingTimeout` and is passed to `post()`.

## Testing
- `yarn install --frozen-lockfile --ignore-scripts`
- `./node_modules/.bin/jest tests/resources/messages/parse.test.ts tests/resources/beta/messages/parse.test.ts --runInBand` → `2 suites passed, 8 tests passed`
- `yarn lint` passed
- `git diff --check`
- Secret scan on diff: clean

### DIFF
--- a/tests/resources/beta/messages/parse.test.ts
+++ b/tests/resources/beta/messages/parse.test.ts
@@ -19,6 +19,27 @@ const messages = new Messages(mockClient);
 describe('Messages.parse()', () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    mockClient._options.timeout = null;
+    mockClient.calculateNonstreamingTimeout.mockReturnValue(600000);
+  });
+
+  it('honors request-level timeout when checking long non-streaming requests', () => {
+    messages.create(
+      {
+        model: 'claude-3-5-sonnet-latest',
+        max_tokens: 100000,
+        messages: [{ role: 'user', content: 'Hello' }],
+      },
+      { timeout: 1200000 },
+    );
+
+    expect(mockClient.calculateNonstreamingTimeout).not.toHaveBeenCalled();
+    expect(mockPost).toHaveBeenCalledWith(
+      '/v1/messages?beta=true',
+      expect.objectContaining({
+        timeout: 1200000,
+      }),
+    );
   });
 
   it('parses structured output correctly', async () => {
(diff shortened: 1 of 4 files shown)


tests/resources/beta/messages/parse.test.ts
tomer-ant, line 1: These tests add to this `parse.test.ts` file (and the other one) aren't really the right place because they don't have anything to do with parsing.

Can they go either in the existing `messages.test.ts` files or in new `timeout.test.ts` files as siblings to these `parse.test.ts` files

tomer-ant:
Thanks, but this is fixed now: https://github.com/anthropics/anthropic-sdk-typescript/issues/1097#issuecomment-5415281519

PR #1110, opened 2026-07-06
fix(aws): preserve AWS options on withOptions() clones so environment helpers work on Claude Platform on AWS

### LINKED ISSUE #1109: AnthropicAws.withOptions() drops AWS options (workspaceId/region/credentials), breaking the self-hosted environment poller/worker on Claude Platform on AWS
### Summary

`AnthropicAws.withOptions()` produces a broken clone: it throws `No workspace ID found` (and, when it doesn't, silently drops the AWS credentials). Because the SDK's self-hosted environment helpers scope a sub-client via `withOptions()`, this makes the environment **poller** and **worker** unusable on Claude Platform on AWS. Any direct `client.withOptions(...)` call is affected too.

### Affected versions

Reproduced on the latest published packages: **`@anthropic-ai/aws-sdk@0.6.0`** with **`@anthropic-ai/sdk@0.110.0`**.

### Root cause

`AnthropicAws`'s constructor destructures the AWS-specific options as named parameters and forwards only the rest (`...opts`) to `super()`:

```ts
constructor({ awsRegion, baseURL, apiKey, awsAccessKey, awsSecretAccessKey,
              awsSessionToken, awsProfile, providerChainResolver, workspaceId,
              skipAuth, ...opts }: AwsClientOptions = {}) {
  ...
  super({ apiKey: resolvedApiKey, baseURL: resolvedBaseURL, ...opts, defaultHeaders: ... });
  ...
}
```

The base `Anthropic` constructor records the options it received on `this._options`, and `withOptions()` rebuilds a clone by spreading `...this._options` back through `new (this.constructor)(...)`. Since the AWS-specific keys were destructured out before `super()`, they never land in `_options` — I verified `'workspaceId' in client._options === false`. The reconstructed clone is therefore missing `workspaceId`, `awsRegion`, and the AWS credentials, so re-validation throws.

### Reproduction

```js
import { AnthropicAws } from '@anthropic-ai/aws-sdk';

const client = new AnthropicAws({
  awsAccessKey: 'AKIAFAKE', awsSecretAccessKey: 'fake',
  awsRegion: 'us-west-2', workspaceId: 'wrkspc_example',
});

// Any clone throws — even a plain header override:
client.withOptions({ defaultHeaders: { 'x-test': '1' } });
// AnthropicError: No workspace ID found. Set `workspaceId` in the constructor
// or the `ANTHROPIC_AWS_WORKSPACE_ID` environment variable.
```

The self-hosted environment poller hits the same failure on its first iteration, because `client.beta.environments.work.poller()` scopes a sub-client via `copyClientForHelper` → `withOptions({ apiKey: null, authToken, credentials: undefined, ... })`:

```js
const it = client.beta.environments.work.poller({
  environmentId: 'env_example', environmentKey: 'unused-on-aws',
});
for await (const _ of it) break;   // throws "No workspace ID found" before the first poll
```

### Expected behavior

`withOptions()` on an AWS client should return a fully-configured `AnthropicAws` clone (workspace id, region, credentials preserved), so the environment poller/worker work on Claude Platform on AWS.

### Workaround

Supplying everything via env vars (`ANTHROPIC_AWS_WORKSPACE_ID`, `AWS_REGION`, and AWS credentials) lets `withOptions()` succeed, since the reconstructed clone re-resolves them from the environment. This defeats explicit constructor configuration.

### Proposed fix

Re-record the AWS-specific options on `this._options` after `super()` so clones round-trip them. PR linked below.

### Note

This is the TypeScript analogue of the Python SDK issue anthropics/anthropic-sdk-python#1736 (there, `AnthropicAWS.copy()` rejected `credentials=`). The TS manifestation is different (lost options rather than a rejected kwarg) but affects the same self-hosted-environment code paths.

### Context

I'm an AWS Specialist Solutions Architect working with the AWS EKS service team. I found this while testing a reference implementation for running Claude Managed Agents self-hosted environments on Amazon EKS as sandboxes, authenticating to Claude Platform on AWS via EKS Pod Identity (SigV4).

csantanapr:
Fixes #1109.

### Problem

`AnthropicAws.withOptions()` produces a broken clone — it throws `No workspace ID found` (or silently drops the AWS credentials). The self-hosted environment helpers (`client.beta.environments.work.poller()`, `EnvironmentWorker`, `SessionToolRunner`) all scope a sub-client via `copyClientForHelper` → `withOptions({ apiKey: null, authToken, credentials: undefined, ... })`, so this makes the environment poller/worker unusable on Claude Platform on AWS. Direct `withOptions()` calls are affected too.

Reproduced on the latest published packages (`@anthropic-ai/aws-sdk@0.6.0` / `@anthropic-ai/sdk@0.110.0`). Full repro in #1109.

### Root cause

The `AnthropicAws` constructor destructures the AWS-specific options as named params and forwards only `...opts` to `super()`. The base client records just what it received on `this._options`, and `withOptions()` rebuilds a clone by spreading `...this._options` back through the constructor — so the clone loses `workspaceId`, `awsRegion`, and the AWS credentials (`'workspaceId' in client._options === false`).

### Fix

Re-record the AWS-specific options on `this._options` after `super()`, so `withOptions()` (and any clone) round-trips the full AWS configuration. Confined to `packages/aws-sdk/src/client.ts`.

### Tests

Added a `withOptions` describe block to `packages/aws-sdk/tests/client.test.ts`:
- plain SigV4 clone preserves workspace id / region / credentials,
- API-key clone preserves workspace id / region,
- the auth-override clone the environment work helpers build preserves workspace id / region.

All three fail without the fix and pass with it. Full `packages/aws-sdk` suite passes (64 tests); `scripts/lint` passes ("Types ok!"); formatter clean.

### Real-world validation

Validated end-to-end using Amazon EKS as the self-hosted environment, with the poller/worker authenticating to Claude Platform on AWS via EKS Pod Identity (SigV4). With the fix, `poller()` scopes its sub-client and polls the work queue successfully instead of throwing on the first iteration.

### Related

TypeScript analogue of the Python SDK fix anthropics/anthropic-sdk-python#1737 (issue #1736).

### DIFF
--- a/packages/aws-sdk/src/client.ts
+++ b/packages/aws-sdk/src/client.ts
@@ -217,6 +217,27 @@ export class AnthropicAws extends Anthropic {
     this.skipAuth = skipAuth;
     this._useSigV4 = resolvedApiKey == null;
 
+    // The base constructor only records the options it received (`apiKey`,
+    // `baseURL`, `...opts`) on `this._options`, which is what `withOptions()`
+    // spreads when it rebuilds a clone. The AWS-specific options above were
+    // destructured out before `super()`, so without this a clone (used by the
+    // environment poller / worker / session-tool-runner, and by any direct
+    // `withOptions()` call) would be reconstructed without them — throwing
+    // "No workspace ID found" and dropping the AWS credentials. Re-record them
+    // so clones round-trip the full AWS configuration.
+    const awsOptions: AwsClientOptions = {
+      ...this._options,
+      awsRegion,
+      awsAccessKey,
+      awsSecretAccessKey,
+      awsSessionToken,
+      awsProfile,
+      providerChainResolver,
+      workspaceId: resolvedWorkspaceId,
+      skipAuth,
+    };
+    this._options = awsOptions;
+
     if (syncRegion || explicitBaseURL || skipAuth) {
       this.ready = Promise.resolve();
     } else {
--- a/packages/aws-sdk/tests/client.test.ts
+++ b/packages/aws-sdk/tests/client.test.ts
@@ -809,4 +809,64 @@ describe('AnthropicAws', () => {
       expect(client.models).toBeDefined();
     });
   });
+
+  describe('withOptions', () => {
+    test('clone preserves AWS SigV4 configuration', () => {
+      const client = new AnthropicAws({
+        awsAccessKey: 'my-access-key',
+        awsSecretAccessKey: 'my-secret-key',
+        awsSessionToken: 'my-session-token',
+        awsRegion: 'us-west-2',
+        workspaceId: 'ws-test',
+      });
+
+      const clone = client.withOptions({ defaultHeaders: { 'x-custom': '1' } });
+
+      expect(clone).toBeInstanceOf(AnthropicAws);
+      expect(clone.workspaceId).toBe('ws-test');
+      expect(clone.awsRegion).toBe('us-west-2');
+      expect(clone.awsAccessKey).toBe('my-access-key');
+      expect(clone.awsSecretAccessKey).toBe('my-secret-key');
+      expect(clone.awsSessionToken).toBe('my-session-token');
+    });
+
+    test('clone preserves API-key configuration', () => {
+      const client = new AnthropicAws({
+        apiKey: 'my-api-key',
+        awsRegion: 'us-east-1',
+        workspaceId: 'ws-test',
+      });
+
+      const clone = client.withOptions({ maxRetries: 5 });
+
+      expect(clone).toBeInstanceOf(AnthropicAws);
+      expect(clone.workspaceId).toBe('ws-test');
+      expect(clone.awsRegion).toBe('us-east-1');
+      expect(clone.maxRetries).toBe(5);
+    });
+
+    test('auth-override clone (as used by the environment work helpers) preserves workspace + region', () => {
+      // Mirrors copyClientForHelper: it calls withOptions({ apiKey: null,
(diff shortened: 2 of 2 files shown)


packages/aws-sdk/src/client.ts
```diff
@@ -217,6 +217,27 @@ export class AnthropicAws extends Anthropic {
...
 
+    // The base constructor only records the options it received (`apiKey`,
+    // `baseURL`, `...opts`) on `this._options`, which is what `withOptions()`
+    // spreads when it rebuilds a clone. The AWS-specific options above were
+    // destructured out before `super()`, so without this a clone (used by the
+    // environment poller / worker / session-tool-runner, and by any direct
+    // `withOptions()` call) would be reconstructed without them — throwing
+    // "No workspace ID found" and dropping the AWS credentials. Re-record them
+    // so clones round-trip the full AWS configuration.
+    const awsOptions: AwsClientOptions = {
```
sophie-ant, line 228: I think there's a bug here.

When we initially construct the AWS client, if an api key isn't passed in, we set `_useSigV4` to true, but we still set the `apiKey` based on the API env var (it's just never used). Now, when we clone, we're passing in the `apiKey` that was resolved from that env var. Because we're passing it explicitly, we set `_useSigV4` to false.

I think once this is fixed we should also add a unit test for it

sophie-ant: I think there's a similar bug with `_baseURLIsExplicit`

sophie-ant (changes requested):
Thanks for your PR! Had a few comments.

sophie-ant:
This is now fixed!

PR #1113, opened 2026-07-07
Remove unsupported idempotency request option

### LINKED ISSUE #1112: idempotencyKey is a silent no-op: base Anthropic client never sets idempotencyHeader
### Summary

`RequestOptions.idempotencyKey` is accepted by the public types and even auto-generated by the client, but the base `Anthropic` client **never sets `idempotencyHeader`**, so the key is silently dropped and no idempotency header is ever sent on a real request.

### Version

`@anthropic-ai/sdk@0.109.0`

### What the types promise

`idempotencyKey?: string` is a documented `RequestOptions` field, so this compiles cleanly and *looks* like it enables idempotency:

```ts
await client.beta.sessions.messages.create(body, { idempotencyKey: "my-key" });
```

### What actually happens

In `src/client.ts` the header is only attached when `this.idempotencyHeader` is truthy:

```ts
// src/client.ts:1457
let idempotencyHeaders: HeadersLike = {};
if (this.idempotencyHeader && method !== 'get') {
  if (!options.idempotencyKey) options.idempotencyKey = this.defaultIdempotencyKey();
  idempotencyHeaders[this.idempotencyHeader] = options.idempotencyKey;
}
```

`idempotencyHeader` is declared `protected idempotencyHeader?: string;` (`src/client.ts:490`) and the base `Anthropic` class never assigns it. So `this.idempotencyHeader` is `undefined`, the `if` is always false, and **no header is emitted** — whether the caller passes `idempotencyKey` or not. `defaultIdempotencyKey()` even returns a value, but it's never used.

### Runtime proof

```js
const { Anthropic } = require('@anthropic-ai/sdk');
const c = new Anthropic({ apiKey: 'sk-test-xxx' });
console.log(c.idempotencyHeader);       // undefined
console.log(c.defaultIdempotencyKey()); // stainless-node-retry-...  (computed but never sent)
```

Confirmed with a request-level trace: the outgoing request carries no idempotency header.

### Impact

The API surface strongly implies passing `idempotencyKey` makes a mutating request (e.g. message/session creation) safe to retry. It does not — the option is a no-op on `Anthropic`, so callers who rely on it to dedupe retried POSTs will still get duplicate side effects. This is easy to misread as "idempotency is supported" from the types alone.

### Expected

One of:

1. The `Anthropic` client sets `idempotencyHeader` to the header the API honors, so `idempotencyKey` actually takes effect; **or**
2. If the Anthropic API genuinely does not support request idempotency, remove/deprecate `idempotencyKey` from the public `RequestOptions` type (or clearly document it as inert) so it doesn't imply a guarantee the client can't deliver.

### Repro

1. `new Anthropic({ apiKey })`
2. Any non-GET call with `{ idempotencyKey: "x" }`
3. Inspect the outgoing request headers — no idempotency header is present.

SakshamKapoor2911:
## Summary

Fixes #1112.

This removes the unsupported `idempotencyKey` request option from the TypeScript SDK surface and deletes the dormant client-side idempotency-header plumbing that depended on an unset `idempotencyHeader`.

I did not add an idempotency header because I could not find an authoritative Anthropic API idempotency header in this repo, the public docs, or the sibling SDKs. Keeping the option in `RequestOptions` makes TypeScript users think the SDK can provide an idempotency guarantee that the client does not actually send.

## Changes

- remove `RequestOptions.idempotencyKey`
- remove the unused `BaseAnthropic.idempotencyHeader` hook
- remove `defaultIdempotencyKey()` and the inactive `buildHeaders` branch
- add a type regression test that `RequestOptions` does not expose `idempotencyKey`

## Validation

- `./node_modules/.bin/jest tests/request-options.test.ts --runInBand`
- `./node_modules/typescript/bin/tsc`
- `./node_modules/.bin/prettier --check src/client.ts src/internal/request-options.ts tests/request-options.test.ts`
- `./node_modules/.bin/eslint src/client.ts src/internal/request-options.ts tests/request-options.test.ts`
- `./scripts/build`
- `./scripts/lint`
- `./node_modules/.bin/jest tests/index.test.ts tests/request-options.test.ts --runInBand`

`./scripts/lint` passed; `publint` emitted the existing package export warnings while exiting successfully.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2,7 +2,6 @@
 
 import type { RequestInit, RequestInfo, BodyInit } from './internal/builtin-types';
 import type { HTTPMethod, PromiseOrValue, MergedRequestInit, FinalizedRequestInit } from './internal/types';
-import { uuid4 } from './internal/utils/uuid';
 import { validatePositiveInteger, isAbsoluteURL, safeJSON } from './internal/utils/values';
 import { sleep } from './internal/utils/sleep';
 export type { Logger, LogLevel } from './internal/utils/log';
@@ -487,7 +486,6 @@ export class BaseAnthropic {
 
   private fetch: Fetch;
   #encoder: Opts.RequestEncoder;
-  protected idempotencyHeader?: string;
   protected _options: ClientOptions;
 
   /**
@@ -846,10 +844,6 @@ export class BaseAnthropic {
 
   private getUserAgent(): string {
     return `${this.constructor.name}/JS ${VERSION}`;
-  }
-
-  protected defaultIdempotencyKey(): string {
-    return `stainless-node-retry-${uuid4()}`;
   }
 
   protected makeStatusError(
@@ -1427,7 +1421,7 @@ export class BaseAnthropic {
     if ('timeout' in options) validatePositiveInteger('timeout', options.timeout);
     options.timeout = options.timeout ?? this.timeout;
     const { bodyHeaders, body } = this.buildBody({ options });
-    const reqHeaders = await this.buildHeaders({ options: inputOptions, method, bodyHeaders, retryCount });
+    const reqHeaders = await this.buildHeaders({ options: inputOptions, bodyHeaders, retryCount });
 
     const req: FinalizedRequestInit = {
       method,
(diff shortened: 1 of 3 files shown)


SakshamKapoor2911:
👋 @jspahrsummers — gentle ping for review. Small fix (16 additions, 3 files), removes unsupported idempotency option from the TypeScript SDK surface. All lint/tests pass. Let me know if anything needs changing.

PR #1118, opened 2026-07-11
fix(tool-runner): forward container id across iterations

### LINKED ISSUE #964: Bug: `toolRunner`'s `setMessagesParams` causes duplicate tool call loops

### LINKED ISSUE #1114: BetaToolRunner does not forward container id, causing 400 when code-execution-backed tools pause for a client tool
### Summary

`BetaToolRunner` never forwards the response's `container.id` on continuation requests. When a server tool that runs on code execution (for example `web_search_20260209` with dynamic filtering) is active and the model pauses for a client tool while the container is live, the runner's next request omits `container` and the API rejects it:

```
HTTP 400 invalid_request_error
"container_id is required when there are pending tool uses generated by code execution with tools."
```

Observed request ids: `req_011Ccoaz3TQSJPcLZssAmSkr` (production hit), `req_011Ccoia6Vj1f7FbaL7gtHtX` (minimal repro below).

### Environment

- `@anthropic-ai/sdk` 0.110.0 (latest at time of filing)
- Node 26, macOS
- Model `claude-opus-4-8`

### Reproduction

A client tool (the memory helper) plus the code-execution-backed web search variant, driven by `toolRunner`. The prompt is engineered so the model records a memory while a search container is open:

```ts
import Anthropic from '@anthropic-ai/sdk'
import { betaMemoryTool } from '@anthropic-ai/sdk/helpers/beta/memory'

const client = new Anthropic({ apiKey: process.env.ANTHROPIC_API_KEY })
const memory = betaMemoryTool({
  view: async () => '- [User location](user-location.md) user is in Lancaster, PA',
  create: async () => 'ok', str_replace: async () => 'ok',
  insert: async () => 'ok', delete: async () => 'ok', rename: async () => 'ok',
})

  model: 'claude-opus-4-8',
  max_tokens: 8000,
  betas: ['context-management-2025-06-27'],
  tools: [memory, { type: 'web_search_20260209', name: 'web_search', max_uses: 5 }],
  messages: [{
    role: 'user',
    content: 'search the web for roofing companies in lancaster pa. also record a memory note that I am getting roofing quotes this month.',
  }],
  stream: true,
})
for await (const stream of runner) { for await (const _ of stream) {} }
await runner.done()
```

Call trace from a failing run:

1. Call 1 (200): model calls the memory client tool.
2. Call 2 (200): response performs web search via code execution (`server_tool_use` blocks for `web_search` and `code_execution`), carries `container: { id: "container_01Q9CzvCUGo5cSz4Harm6ngG", ... }`, and stops with `stop_reason: "tool_use"` on two pending `memory` tool_use blocks issued while the container is live.
3. Call 3 (400): the runner appends the memory tool results and re-sends. The request body has no `container` key at all, and the API rejects it with the error above.

`grep -c container lib/tools/BetaToolRunner.mjs` on 0.110.0 returns 0; the runner neither captures nor forwards the id.

Notes from control experiments: replaying fully completed code-execution blocks across turns without `container` is accepted (a fresh container is spawned), and a synthetic orphaned pending `code_execution` `server_tool_use` alone is also accepted. The 400 specifically requires a client tool_use generated while the container is open, answered without the id.

### Expected behavior

The runner should capture `container.id` from each response and include `container` on its continuation requests, the same way callers must when hand-rolling the loop.

### Workaround

Carrying the id forward manually between iterations works and is verified end to end:

```ts
for await (const stream of runner) {
  for await (const event of stream) { /* ... */ }
  const final = await stream.finalMessage()
  if (final.container?.id) {
    (runner.params as { container?: string | null }).container = final.container.id
  }
}
```

Note `setMessagesParams` is not usable for this: it flips the runner's internal mutated flag, which skips pushing the assistant turn into history and re-sends the same request.

AKnassa:
Fixes #1114. Fixes #964.

Supersedes #966 by @jeanpierrecarvalho, which has been the consolidation point for this bug since March but has been conflicting with `main` since April. This PR keeps both of its ideas and closes three gaps in it — details below, and I'm happy to hand this to @jeanpierrecarvalho or close it if they'd rather rebase theirs.

## What this does

When Claude uses a tool that runs in a container (web search with filtering, code execution) and then pauses to call *your* code, the container is still alive and holding that paused work. The API needs the container's id on the next request to pick it back up.

The tool runner never remembered that id, so the next request went out without it and the API rejected the run:

```
400 invalid_request_error
container_id is required when there are pending tool uses generated by code execution with tools
```

Now the runner remembers the id and sends it back. The run finishes instead of dying halfway through.

## Why this shape, and not the earlier ones

Five PRs have gone at this (#966, #1073, and #974/#981/#994 which were closed as duplicates). They all reached for the same one-liner, and it has three holes worth naming, because the holes are why the bug is still shipping:

**1. It skips anyone who already set `container`.** The guard used is:

```ts
if (message.container?.id != null && this.#state.params.container == null) {
  this.#state.params.container = message.container.id;
}
```

`container` accepts an object, not just an id — `BetaContainerParams = { id?, skills? }`. So a caller who passes `container: { skills: [...] }` to preload skills is *not* `== null`, the id is never captured, and they keep hitting the exact 400 this is meant to fix. This PR fills in only the id and leaves the rest of their config alone.

**2. It captures once and never refreshes.** Because it only writes when `params.container == null`, a container minted later in the run is never picked up, and if the server mints a *new* one the runner keeps sending the old, now-stale id forever.

**3. It captures inside the compaction branch,** so on a compaction turn the id is never recorded at all.

There's also a fourth thing, which is the real reason this PR is bigger than a one-liner: **storing the id in `params` doesn't survive the API's own escape hatch.** `setMessagesParams()` replaces the params object wholesale, so a container id living in there is silently dropped by any caller who uses it — including the workaround people are passing around for this very bug. The id is runner state, not caller config, so it lives in a private field.

## The other half of #964

#964 has a second symptom in its title — *"`setMessagesParams` causes duplicate tool call loops"* — and it has the same root. One flag, `#mutated`, was doing two different jobs. Any call to `setMessagesParams` set it, even one that just bumped `max_tokens`. That was read as *"the caller has taken over the conversation history"*, so the runner threw away the assistant turn it had just received and re-sent the same request — and the tools ran again.

Whether the caller took over is now decided by looking at whether `messages` actually changed (by identity, so an in-place edit counts too), not by "did anything at all change". `#mutated` is gone; both decisions come from that one check, so they can't drift apart again.

## What changed

- The runner records the container id from each response and fills it in on the next request.
- A container **you** set always wins. An id of your own, or an explicit `null` meaning you want none, is passed through untouched. Only what you left unset gets filled in — so `{ skills: [...] }` keeps its skills and just gains an `id`.
- A response that reports no container doesn't wipe a remembered one. A turn that runs no code reports `container: null` while the container it made is still very much alive.
- Changing `max_tokens` (or any non-`messages` param) no longer makes the runner drop the assistant turn and re-run your tools.

Two files, both hand-written (`src/lib/`), nothing generated. The change actually *removes* a class field rather than adding one.

## One deliberate behavior change, called out

A param-only `setMessagesParams` on a turn with **no tool calls** now ends the run instead of firing a second, near-identical request. That second request was the duplicate-tool-call loop, so it can't stay. If you want another turn, `pushMessages(...)` still does it.

I tried keeping the old loop-exit and only fixing the history gate — it re-opens the loop (6 requests where there should be 2). They have to move together.

## How to check it

`yarn test` — `tests/lib` is 355 green across 19 suites, including the recorded E2E cassette **unchanged**, which is itself the proof that a run with no container puts the exact same bytes on the wire as before.

19 new tests, all asserting on the **outgoing request body** and counting tool invocations, because internal state can look right while the wire is wrong. They cover: the issue's 3-call trace (streaming and non-streaming), the `skills` object, a refreshed id, a `null` response, a caller-set id, compaction, `pushMessages` mid-turn, wholesale param replacement, and in-place mutation of the history.

Reproducing the issue's exact trace, before and after:

```
BEFORE                                  AFTER
call 1: container = <absent>            call 1: container = <absent>
call 2: container = <absent>            call 2: container = <absent>
call 3: container = <absent>  -> 400    call 3: container = "container_01Q9…"  -> ok
```

## For a maintainer

- **Worth one live-fire check:** the fix can send `container: { skills: [...], id: "container_…" }`. It's type-valid and no mocked test can catch a 400 on that shape. If `id` and `skills` are mutually exclusive on the wire, say so and I'll change it.
- `#checkAndCompact()` builds its own summarization request and doesn't forward the container. Harmless today (it strips pending `tool_use` blocks first, so the 400 can't trigger) and `compactionControl` is deprecated — flagging it, not fixing it here.
- The non-beta `src/lib/tools/ToolRunner.ts` still has the old gate. It has zero importers, so I left it alone rather than widen this diff — happy to port it or delete it if you'd like.
- Known limit: mutating a message object's *interior* (`Object.assign(params.messages[0], …)`) is invisible to any identity check. Rebuilding the array is the supported way, and it's documented on `setMessagesParams` now.


tomer-ant:
Thanks, but this is now fixed: https://github.com/anthropics/anthropic-sdk-typescript/issues/1114#issuecomment-5415100070

PR #1129, opened 2026-07-22
fix(client): release abort listeners when requests settle

codebytere-ant:
The client adds a `{ once: true }` abort listener to a caller-provided signal for every HTTP attempt (`fetchWithTimeout`) and never removes it. A long-lived signal - one AbortController reused across a session - accumulates one listener plus its bound per-attempt AbortController per request until the signal fires or is collected, and Node prints `MaxListenersExceededWarning` at the 11th. The bound `abort` already keeps each entry's retained graph small; the count is what grows without bound.

The listener can't be removed when fetch resolves - it has to survive until the body settles, or aborting an in-flight body read stops propagating. So the removal is registered in a WeakMap keyed by the per-request controller (`src/internal/request-signal.ts`) and released at the points that already own the end of a request: buffered-body parsing (`internal/parse.ts`), stream teardown (both SSE iterators' existing `finally` in `core/streaming.ts`), the retry path (right after the body cancel), and the connection-error and status-error paths in `makeRequest`. A raw `Response` handed back to the caller (`__binaryResponse`, `.asResponse()`) keeps its listener so aborting an in-flight download still works - the change is strictly a release, never an earlier cutoff.

`tests/abort-listener-cleanup.test.ts` pins it: settled, failed, connection-error, and retried requests plus consumed and early-broken streams all leave zero listeners on the caller signal (six of the seven fail before this change), and abort still cancels an in-flight request. For what it's worth the same pattern is in the other Stainless SDKs - happy to point this at the generator instead if that's the better home.

Update: added a FinalizationRegistry backstop for responses that never reach a release point - a partially iterated stream or a raw binary response whose reference is dropped. It holds only the request controller (never the response), and explicit release unregisters it, so deterministic cleanup stays primary. Feature-detected; every supported runtime has it (ES2021). New test: abandon a half-iterated stream, listener is released after collection - fails without the backstop.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1205,7 +1205,7 @@ export class BaseAnthropic {
       }),
     );
 
-    armAbandonmentBackstop(response, controller);
+    armAbandonmentBackstop(response.body ?? response, controller);
     return { response, options, controller, requestLogID, retryOfRequestLogID, startTime };
   }
 
@@ -1251,7 +1251,7 @@ export class BaseAnthropic {
     const abort = this._makeAbort(controller);
     if (signal) {
       signal.addEventListener('abort', abort, { once: true });
-      registerRequestSignalCleanup(controller, () => signal.removeEventListener('abort', abort));
+      registerRequestSignalCleanup(controller, signal, abort);
     }
 
     const isReadableBody =
--- a/src/client.ts
+++ b/src/client.ts
@@ -9,6 +9,7 @@ export type { Logger, LogLevel } from './internal/utils/log';
 import { castToError, isAbortError } from './internal/errors';
 import type { APIResponseProps } from './internal/parse';
 import { getPlatformHeaders } from './internal/detect-platform';
+import { registerRequestSignalCleanup, releaseRequestSignal } from './internal/request-signal';
 import * as Shims from './internal/shims';
 import * as Opts from './internal/request-options';
 import { stringifyQuery } from './internal/utils/query';
@@ -1057,6 +1058,7 @@ export class BaseAnthropic {
     const headersTime = Date.now();
 
     if (response instanceof globalThis.Error) {
+      releaseRequestSignal(controller);
       const retryMessage = `retrying, ${retriesRemaining} attempts remaining`;
       if (options.signal?.aborted) {
         throw new Errors.APIUserAbortError();
@@ -1142,6 +1144,7 @@ export class BaseAnthropic {
 
         // We don't need the body of this response.
         await Shims.CancelReadableStream(response.body);
+        releaseRequestSignal(controller);
         loggerFor(this).info(`${responseInfo} - ${retryMessage}`);
         loggerFor(this).debug(
           `[${requestLogID}] response error (${retryMessage})`,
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -1240,7 +1249,10 @@ export class BaseAnthropic {
     // the lifetime of the signal. Using `.bind()` only retains a reference to the
     // controller itself.
     const abort = this._makeAbort(controller);
-    if (signal) signal.addEventListener('abort', abort, { once: true });
+    if (signal) {
+      signal.addEventListener('abort', abort, { once: true });
+      registerRequestSignalCleanup(controller, () => signal.removeEventListener('abort', abort));
```
craigie-ant: note you should not use arrow functions here because it captures all local variables - including the request - so if we miss cleaning it up later then we're leaking request bodies which can be huge! this should use a pattern like `makeAbort()` does right above.

craigie-ant (changes requested):
Thanks for catching this Shelley!

I think there's a bug with the streaming response logic where you can hold a reference to the response body but *not* the response, then because the `Response` wrapper is GCd we remove the listener - even though it could still fire because its attached to the body stream. Does that make sense?

I think we could fix this by registering the GC listener on `response.body` instead of `response` but not super confident.

<details>
<summary>Claude comment</summary>

The GC backstop is an earlier cutoff, not just a release

The finalizer is keyed on the `Response` wrapper, but the wrapper is collectable while its body is still being read. `messages.batches.results()` hits this directly: it's `stream: true` + `__binaryResponse: true`, `defaultParseResponse`'s `.finally` skips it, and the returned `JSONLDecoder` holds only `response.body`'s async iterator — so once the `APIPromise` resolves, nothing references `props.response`.

Reproduced with the real global fetch (undici) against a slow local JSONL server:

- read a few rows from the decoder, let a GC happen → the caller signal's listener count goes 1 → 0 mid-iteration
- `signal.abort()` afterwards does nothing — rows keep streaming, `decoder.controller.signal.aborted === false`
- on `main` the same probe aborts the read with `AbortError`, as expected

So for that path (and any caller that keeps `res.body` but drops the `Response`) this *is* an earlier cutoff, contra the PR description. Options: give `JSONLDecoder` a `finally`-based release like the SSE iterators and stop arming the backstop for `stream`/`__binaryResponse` responses, or drop the backstop entirely and document that raw/binary responses keep their listener until abort/GC.

</details>

src/client.ts
codebytere-ant: good catch - moved the closure into a module-scope factory in request-signal.ts so it captures exactly the signal and the listener, same shape as makeAbort. fixed in f29baba

codebytere-ant:
yes that makes sense - one refinement: the built-in Stream can't hit it (the SSE iterator closes over the Response, so it stays alive while the stream is consumed), but .asResponse() callers can keep a body reader and drop the wrapper, and then the old target released the listener mid-read exactly as you describe. registered on response.body as you suggested (falling back to the Response when body is null) in f29baba, with a test pinning the scenario: reader held through GC keeps the listener armed, dropping the reader releases it - it fails against the old target.

craigie-ant (approved):
Thanks!

PR #1131, opened 2026-07-23
fix(streaming): await async event listener callbacks before resolving done()

### LINKED ISSUE #775: add capability to await async function for on event handler
In `anthropic.messages.stream` function call, you can subscribe to events with `on` method. the callback(handler) for this function can not be async(for example sending an http request in the callback with true await capability). I checked the source code, inside `_emit` method in `MessageStream.ts` file there is something like:
```   
if (listeners) {
      this.#listeners[event] = listeners.filter((l) => !l.once) as any;
      listeners.forEach(({ listener }: any) => listener(...args));
    }
```
I think it should be `await listner(...args)` so callback can be awaited on asynchronous handlers. Please let me know what you think.

okxint:
Fixes #775.

`stream.on('message', async (msg) => { ... })` and similar async handlers are currently fire-and-forget. The internal `_emit` method calls `listeners.forEach(({ listener }) => listener(...args))` — synchronous, no `await` — so Promises returned by async handlers are never tracked. `done()` (and `await stream`) resolves immediately after the `'end'` event fires, before any async listeners complete.

The fix adds a `#listenerAwaitQueue` private field to both `MessageStream` and `BetaMessageStream`. When `_emit` runs listeners, any returned `Promise` is pushed into the queue. When the `'end'` event fires, resolution of the end promise is deferred until all queued Promises settle via `Promise.all`. Sync handlers are unaffected — they return `undefined`, which is not pushed to the queue. Errors from async listeners resolve the end promise rather than leaving it pending (the unhandled rejection from the listener propagates normally to the user).

### DIFF
--- a/src/lib/BetaMessageStream.ts
+++ b/src/lib/BetaMessageStream.ts
@@ -70,6 +70,7 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
 
   #ended = false;
   #errored = false;
+  #listenerAwaitQueue: Promise<unknown>[] = [];
   #aborted = false;
   #catchingPromiseCreated = false;
   #response: Response | null | undefined;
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Async listener rejection is currently swallowed: `Promise.all(queue).then(settle, settle)` resolves the end promise on either outcome. A synchronous listener exception can fail the stream, but the equivalent async listener rejection now lets `done()` succeed, hiding the callback failure. Should the rejection path reject the end promise / enter the normal stream error path instead?

Issue #1134, opened 2026-07-24
Use predictable user-agent string for requests


rwalle:
https://github.com/anthropics/anthropic-sdk-typescript/blob/9ddb1adbea8f748394dd4b8dd0c7e9a38d5cff9b/src/client.ts#L790-L792

Here `this.constructor.name` could be a mangled string after JavaScript build process (e.g. minifier), resulting in unrecognizable client when inspecting network traffic.

This is the root cause of https://github.com/anthropics/claude-code/issues/80191.

(Thanks to @nemo-lm for tracing it to SDK code)

patrickswedish:
I can take this. I’ll replace the constructor-name-derived User-Agent token with a stable SDK identifier and add focused regression coverage proving minification/subclassing cannot change the emitted identity.

dtmeadows-ant:
@rwalle thanks for the report — fixed in sdk-v0.116.0: https://github.com/anthropics/anthropic-sdk-typescript/releases/tag/sdk-v0.116.0. The User-Agent is now a hardcoded `Anthropic/JS <version>` string (and likewise for the Bedrock/Vertex/Foundry clients), so minification or subclassing no longer changes it.

PR #1136, opened 2026-07-27
fix: replace constructor.name with literal string in getUserAgent

okxint:
When the SDK is bundled with a minifier, `this.constructor.name` gets mangled to a single-character identifier (`r`, `e`, etc.), producing user-agent strings like `Anthropic-JS/r 0.x.x` instead of `Anthropic-JS/Anthropic 0.x.x`.

Replace the runtime-computed name with the literal string `'Anthropic'`, which is always stable after minification.

Related: anthropics/anthropic-sdk-typescript#1134

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -863,7 +863,7 @@ export class BaseAnthropic {
   }
 
   private getUserAgent(): string {
-    return `${this.constructor.name}/JS ${VERSION}`;
+    return `Anthropic/JS ${VERSION}`;
   }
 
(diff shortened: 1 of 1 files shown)


dtmeadows-ant:
@okxint thanks — this is essentially what shipped in sdk-v0.116.0 (https://github.com/anthropics/anthropic-sdk-typescript/releases/tag/sdk-v0.116.0) for #1134, so closing this out.

PR #1145, opened 2026-08-07
docs(helpers): add missing .beta segment in toolRunner heading

MsfPablo:
The `toolRunner` method is only defined on the beta Messages resource (`src/resources/beta/messages/messages.ts`), not on the non-beta Messages resource. Every runnable example in `helpers.md` already calls `anthropic.beta.messages.toolRunner(...)`, so the heading at line 359 was missing the `.beta` segment.

This change updates the heading to match the actual API path.

I used Claude Code to assist with this one-line docs fix and reviewed the diff manually.

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -356,7 +356,7 @@ const calculatorTool = betaTool({
 });
 ```
 
-### `client.messages.toolRunner(params): BetaToolRunner`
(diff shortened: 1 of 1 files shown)


MsfPablo:
Small docs ping — the heading is missing the `.beta` segment, so the snippet underneath doesn't match the actual call path.

PR #1150, opened 2026-08-13
docs: fix toolRunner heading to include .beta segment

### LINKED ISSUE #1141: helpers.md: tool runner method heading is missing the .beta segment
Thanks for the tool runner helpers, they are handy.

While reading helpers.md I noticed the method reference heading drops the `beta` segment, so the path as written does not resolve on the client.

helpers.md line 359:

```
### `client.messages.toolRunner(params): BetaToolRunner`
```

`toolRunner` is only defined on the beta Messages resource (`src/resources/beta/messages/messages.ts`). The non-beta `src/resources/messages/messages.ts` exposes `create`, `parse`, `stream`, and `countTokens`, but no `toolRunner`. Every runnable example in the same file calls `anthropic.beta.messages.toolRunner(...)` (lines 199, 215, 237, 266), so the heading looks like a dropped `.beta`.

Suggested fix:

```
### `client.beta.messages.toolRunner(params): BetaToolRunner`
```

I checked the open docs PR #881 and this particular heading looks untouched by it, so flagging separately. Happy to open a PR instead if you prefer.

juneja-varun:
`toolRunner` is only defined on the beta Messages resource (`src/resources/beta/messages/messages.ts`). The non-beta `src/resources/messages/messages.ts` exposes `create`/`parse`/`stream`/`countTokens`, but no `toolRunner`. Every runnable example already below this heading calls `anthropic.beta.messages.toolRunner(...)` — only the heading itself was missing the `.beta` segment, so the path as written doesn't resolve on the client.

One-line fix.

Fixes #1141

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -356,7 +356,7 @@ const calculatorTool = betaTool({
 });
 ```
 
-### `client.messages.toolRunner(params): BetaToolRunner`
+### `client.beta.messages.toolRunner(params): BetaToolRunner`
 
 **Parameters:** All standard message parameters plus:
 


juneja-varun:
Checking in here — small docs fix, happy to adjust if you'd like it done differently.

PR #1153, opened 2026-08-14
fix: preserve zero retry-after-ms precedence

jstar0:
Summary:
- Preserve `retry-after-ms` precedence when the parsed delay is 0.
- Add regression coverage for responses that send both retry headers.

Reason:
`retry-after-ms` is parsed before `Retry-After`, but the follow-up guard treated a valid 0 ms delay as absent. When both headers were present, `Retry-After` could replace the more precise millisecond delay.

Verification:
- `pnpm exec jest tests/index.test.ts -t "retry-after-ms zero takes precedence over retry-after" --runInBand`
- `pnpm exec jest tests/index.test.ts -t "retries" --runInBand`
- `pnpm lint`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1388,7 +1388,7 @@ export class BaseAnthropic {
 
     // About the Retry-After header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After
     const retryAfterHeader = responseHeaders?.get('retry-after');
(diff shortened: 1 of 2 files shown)


jstar0:
Re-ran the focused `retry-after-ms` precedence regression and `pnpm lint` on this branch; both still pass.

PR #1154, opened 2026-08-14
fix: handle Response uploads without URLs

jstar0:
Summary:
- Let `toFile()` handle `Response` objects whose `url` is empty.
- Add regression coverage for the default filename fallback.

Reason:
The Fetch standard permits manually constructed `Response` objects to have an empty `url`. `toFile(new Response(...))` currently calls `new URL(value.url)` in the Response path and throws before it can fall back to `unknown_file`.

Verification:
- `pnpm exec jest tests/uploads.test.ts -t "Response without a URL" --runInBand`
- `pnpm exec jest tests/uploads.test.ts --runInBand`
- `pnpm lint`

### DIFF
--- a/src/internal/to-file.ts
+++ b/src/internal/to-file.ts
@@ -107,7 +107,9 @@ export async function toFile(
 
   if (isResponseLike(value)) {
     const blob = await value.blob();
-    name ||= new URL(value.url).pathname.split(/[\\/]/).pop();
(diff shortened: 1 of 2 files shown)


jstar0:
Re-ran the `Response without a URL` regression and `pnpm lint` on this branch; both still pass.

PR #1160, opened 2026-08-23
fix: use client logger for stream parse errors

### LINKED ISSUE #1159: defaultParseResponse bypasses the configured logger for SSE parse errors
### Summary

`defaultParseResponse` receives the active `BaseAnthropic` client, but the default streaming branch does not pass it to `Stream.fromSSEResponse`:

```ts
return Stream.fromSSEResponse(response, props.controller) as any;
```

`Stream.fromSSEResponse` already accepts an optional third `client` argument. When it is omitted, the stream parser falls back to the global `console` instead of using `loggerFor(client)`:

```ts
const logger = client ? loggerFor(client) : console;
```

Consequently, a client configured with `logLevel: "off"` or a custom logger can still write malformed SSE payload data and raw chunks directly to `console.error` when JSON parsing fails.

### Expected behavior

The default response parser should pass its existing client to the stream:

```ts
return Stream.fromSSEResponse(response, props.controller, client) as any;
```

This makes streaming parse diagnostics respect the client's configured logger and log level. Error propagation remains unchanged.

### Affected code

- `src/internal/parse.ts`: default streaming response construction
- `src/core/streaming.ts`: logger selection and malformed-JSON diagnostics

### Version

Observed in `@anthropic-ai/sdk@0.97.1`. The same omission is present on the current `main` branch at the time of filing.

dawNotPoi:
## Summary
- pass the active client to the SSE stream parser
- add regression coverage that malformed SSE diagnostics use the configured logger instead of the global console

## Testing
- corepack pnpm exec jest tests/index.test.ts --runInBand
- corepack pnpm lint

Fixes #1159

### DIFF
--- a/src/internal/parse.ts
+++ b/src/internal/parse.ts
@@ -28,7 +28,7 @@ export async function defaultParseResponse<T>(
       // Note: there is an invariant here that isn't represented in the type system
       // that if you set `stream: true` the response type must also be `Stream<T>`
 
-      return Stream.fromSSEResponse(response, props.controller) as any;
+      return Stream.fromSSEResponse(response, props.controller, client) as any;
     }
 
     // fetch refuses to read the body when the status code is 204.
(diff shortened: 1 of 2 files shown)


craigie-ant:
thanks for the PR! sorry I didn't see this before we fixed this separately.

PR #1168, opened 2026-08-26
fix(credentials): stop superseded token refreshes from committing stale results

latent-9:
## What

`TokenCache` guards against a forced refresh coalescing into an in-flight non-forced one, but the non-forced refresh's own completion handler still unconditionally committed `this.cached`. A slower superseded refresh could therefore overwrite the newer token after it had already been cached.

## Why

Concrete interleaving with the current code:

1. Request A triggers a plain refresh (call 1, in flight).
2. A concurrent request gets a 401 → `invalidate()` bumps state, or request B forces a refresh (call 2).
3. Call 1 settles **after** call 2 and its completion handler writes its (older, disk-cached) token into `this.cached`.

The next `getToken()` then serves the stale token that was just rejected with a 401, repeating the loop. The existing doc comment on `refresh()` already identifies half of this hazard ("may re-serve the same stale disk token") but only the coalescing side was guarded.

## Change

Add a monotonically increasing `generation` counter:

- `invalidate()` and every new `doRefresh()` supersede older generations.
- Only the latest generation may commit `cached` / clear `pendingRefresh`; superseded completions return their result to their own awaiter without touching shared state.

Two regression tests cover the forced-overwrite race and the invalidated-mid-flight case; both fail on main and pass here.

### DIFF
--- a/src/lib/credentials/token-cache.ts
+++ b/src/lib/credentials/token-cache.ts
@@ -27,6 +27,10 @@ export class TokenCache {
   private cached: AccessToken | null = null;
   private pendingRefresh: Promise<AccessToken> | null = null;
   private nextForce = false;
+  /** Increments on every new refresh and on invalidate(), so a slower
+   *  superseded refresh can no longer commit its (older) token over a
+   *  newer one. */
+  private generation = 0;
   private lastAdvisoryError = 0;
   private onAdvisoryRefreshError: ((err: unknown) => void) | undefined;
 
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The generation check prevents a superseded refresh from updating `cached`, but the caller awaiting that old refresh still receives its token because the mismatch branch returns `token`. In the new forced-refresh regression, `first` can therefore resolve to `"stale"` after `invalidate()`. Should superseded callers join or retry the current generation instead of returning a token already known to be superseded?

PR #1171, opened 2026-08-27
fix: prevent retries with streaming request bodies

### LINKED ISSUE #1170: Retrying a request with a single-use stream body silently sends a 0-byte body and reports success
### Describe the bug

When a request body is a single-use async iterable (or ReadableStream) and the first attempt fails with a retryable error (500/429/408/409/connection failure), the retry re-wraps the same exhausted iterator: the second attempt goes on the wire with a **0-byte body**, and the caller's promise **resolves successfully**. The payload is silently gone — no error, no warning.

In `buildBody` (`src/client.ts`), an async-iterable body is wrapped via `ReadableStreamFrom(options.body)` per attempt, but `options.body` itself is the caller's single-use iterator: attempt 1 drains it, and the retry gets an empty stream.

### Steps to reproduce

Self-contained script against `@anthropic-ai/sdk@0.121.0` — a loopback server fails the first POST with 500 and accepts the second; the client streams a 1024-byte body with `maxRetries: 1`:

<details>
<summary>repro (plain node, no dependencies beyond the SDK)</summary>

```js
// repro: node repro.mjs — loopback only, no external traffic
import http from 'node:http';
import { Anthropic } from '@anthropic-ai/sdk';

let n = 0;
const bodies = [];
const server = http.createServer((req, res) => {
  const chunks = [];
  req.on('data', (c) => chunks.push(c));
  req.on('end', () => {
    n++;
    bodies.push(Buffer.concat(chunks).toString('utf8'));
    const body = n === 1 ? JSON.stringify({ error: { type: 'api_error' } }) : JSON.stringify({ id: 'msg_ok', type: 'message' });
    res.writeHead(n === 1 ? 500 : 200, { 'content-type': 'application/json', 'content-length': Buffer.byteLength(body) });
    res.end(body);
  });
});
await new Promise((r) => server.listen(0, '127.0.0.1', r));

const client = new Anthropic({ apiKey: 'repro-local', baseURL: `http://127.0.0.1:${server.address().port}`, maxRetries: 1 });
async function* bodyIter() { yield 'A'.repeat(512); yield 'A'.repeat(512); } // 1024 bytes, single-use

const res = await client.post('/v1/anything', { body: bodyIter(), headers: { 'content-type': 'application/octet-stream' } });
console.log('attempts:', n, '| body lengths sent:', bodies.map((b) => Buffer.byteLength(b)));
console.log('caller sees success:', res && res.id === 'msg_ok');
server.close();
```

</details>

Output:

```
attempts: 2 | body lengths sent: [ 1024, 0 ]
caller sees success: true
```

### Expected behavior

Either the retry carries the same body as the original, or the retry is refused with a clear error when the body source is single-use. A silently-empty request that reports success is the worst of the three outcomes — the server may act on the truncated request (e.g. create an empty object) while the application believes its full payload was delivered.

### Environment

- `@anthropic-ai/sdk` 0.121.0
- Node 24 (also applies wherever streaming bodies are supported)

### Suggested fix

Buffer stream bodies up to a size cap for retry (as JSON bodies effectively are), or track that the body source is single-use and fail the retry loudly instead of sending an empty one.

xianjianlf2:
## Summary
Prevent retries from re-sending a consumed async iterable or `ReadableStream` request body. Requests with single-use bodies make one attempt and preserve the original API or connection error, including its status, response payload, or cause. Replayable request bodies retain the existing retry behavior.

Merge the current upstream main, retaining upstream retry-count validation and the single-use body regressions.

Fixes #1170.

## Validation
- `pnpm test tests/index.test.ts --runInBand`: 60 main-package tests passed; all five SDK subpackage test suites passed (3 existing skipped tests).
- `pnpm lint`: Prettier, ESLint, build, type checks, Are The Types Wrong, and publint passed.
- `git diff --check` passed.

AI assistance: OpenAI Codex.

## Validation update (2026-09-22)
Merged upstream main, adapting the retry guards to debugLogRequestDetails. `pnpm exec vitest run tests/index.test.ts`: 64 passed. `pnpm lint`: formatting, ESLint, build, type checking, Are The Types Wrong and publint passed. Head: f83f864.


sylvesterkaczmarek (commented):
Could this skip the retry while preserving the original API or transport error? The check runs only after the first failed request reaches `retryRequest()`, so a streaming-body request that receives a useful 500 or 429 is replaced by `AnthropicError("Cannot retry...")`. That makes the non-replayability detail mask the server failure that actually caused the retry decision.

xianjianlf2:
Addressed the review feedback in ab5b3f5f:

- streaming request bodies now skip retries in the retry-decision paths
- the original HTTP API error or transport error is preserved
- added regression coverage for async iterable and ReadableStream HTTP failures, plus a ReadableStream connection failure
- rebased onto the latest main

Validation: targeted regression tests pass; pnpm lint (format, ESLint, build, types, package checks) passes; all five SDK subpackage suites pass.

xianjianlf2:
Synced this branch with the latest `main` (17 upstream commits). The only overlapping file was `src/client.ts`; the upstream workspace-ID exports / generated comment formatting and this PR’s single-use-body retry behavior touch separate sections, so both intents are preserved. The PR diff remains limited to `src/client.ts` and `tests/index.test.ts`.

Validation after the merge:
- `pnpm test tests/index.test.ts --runInBand` — main suite 54/54 passing; all five SDK subpackage suites also passed (with 3 pre-existing skipped Google Cloud tests)
- `pnpm lint` — Prettier, ESLint, build, type checks, ATTW, and publint passed
- `git diff --check upstream/main...HEAD` — passed

sylvesterkaczmarek (commented):
Re-reviewed current `5edffb8`. The retry gate now consistently excludes single-use bodies on both connection failures and retryable HTTP responses. `ReadableStream`, async iterables, and iterator objects are treated conservatively as non-replayable, while ordinary replayable bodies keep the existing retry path. The tests verify one attempt for both streaming forms and preserve the original connection error. No blocker from me on this head.

sylvesterkaczmarek (commented):
Rechecked current `9a0c835e` after the latest main sync. The single-use-body retry gate is still present on both connection failures and retryable HTTP responses, and the async-iterable, ReadableStream, and preserved-connection-error regressions remain intact. No new blocker from me after the sync.

sylvesterkaczmarek (approved):
Rechecked current f83f864e after the latest upstream sync. The single-use-body retry guard remains on both connection failures and retryable HTTP responses, ReadableStream/async-iterable bodies still make only one attempt, and the regressions preserve the original API or connection error rather than replacing it with a replayability error. No blocker from me.

Issue #1162, opened 2026-08-24
transformJSONSchema drops $defs when the schema root is a $ref


cmun2:
### Summary

`transformJSONSchema` drops `$defs` when the schema root is a `$ref`, leaving the reference dangling.

`$ref` is popped and returned before `$defs` is ever read ([`src/lib/transform-json-schema.ts#L31-L44`](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/lib/transform-json-schema.ts#L31-L44)):

```ts
const ref = pop(jsonSchema, '$ref');
if (ref !== undefined) {
  strictSchema['$ref'] = ref;
  return strictSchema;          // returns before the $defs block below
}

const defs = pop(jsonSchema, '$defs');
```

### Reproduction

No API key, no network — this is local schema normalisation.

```console
$ npm i @anthropic-ai/sdk@0.120.0
```

```js
import { transformJSONSchema } from '@anthropic-ai/sdk/lib/transform-json-schema';

transformJSONSchema({
  $ref: '#/$defs/Item',
  $defs: { Item: { type: 'object', properties: { a: { type: 'string' } } } },
});
```

```
{ "$ref": "#/$defs/Item" }
```

The definitions are gone, so `#/$defs/Item` resolves to nothing.

### Why this shape shows up

It is what pydantic's `RootModel` and `zod-to-json-schema` emit for a top-level model — the root is a `$ref` and the definitions sit beside it. A `$ref` nested inside `properties` is unaffected; only the root case is.

### The Python SDK already fixed this

Same defect, same function, fixed in [anthropics/anthropic-sdk-python#1642](https://github.com/anthropics/anthropic-sdk-python/pull/1642) (merged 2026-06-04) by moving `$defs` processing above the `$ref` early-return. The Python source still carries the comment explaining why the order matters.

Same input, the two official SDKs today:

| | Python 1.0.0 | TypeScript 0.120.0 |
|---|---|---|
| `{"$ref": "#/$defs/Item", "$defs": {...}}` | definitions preserved | **definitions dropped** |

### Environment

- `@anthropic-ai/sdk@0.120.0` (published package, verified directly)
- also present on `main` at `bfa9197`
- Node 23

PR attached.

PR #1174, opened 2026-08-29
fix: preserve defs when schema root is a ref, and fix compaction delta null content

AbhiPra24:
## Summary

1. **Issue #1162**: `transformJSONSchema` previously performed an early return when encountering a `$ref`, which caused top-level `$defs` and `definitions` to be discarded when the schema root was a `$ref`. Reordered processing so `$defs` and `definitions` are transformed and retained on the returned strict schema.
2. **Issue #1164**: `BetaMessageStream` previously coerced `null` `content` in `compaction_delta` events into the string `"null"` via string concatenation (`'' + null`). Fixed delta accumulation to preserve `null` content properly and keep `encrypted_content` intact across deltas.

## Test Plan
- Added unit tests in `tests/helpers/transform-json-schema.test.ts` verifying `$defs` and `definitions` preservation when root schema is a `$ref`.
- Added unit tests in `tests/api-resources/BetaMessageStream.test.ts` verifying `compaction_delta` null content handling and non-null accumulation.
- Ran `npm run lint` and verified test suite passes.

### DIFF
--- a/src/lib/BetaMessageStream.ts
+++ b/src/lib/BetaMessageStream.ts
@@ -495,7 +495,7 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
             break;
           }
           case 'compaction_delta': {
-            if (content.type === 'compaction' && content.content) {
+            if (content.type === 'compaction' && content.content && event.delta.content !== null) {
               this._emit('compaction', content.content);
             }
             break;
--- a/src/lib/BetaMessageStream.ts
+++ b/src/lib/BetaMessageStream.ts
@@ -495,7 +495,7 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
             break;
           }
           case 'compaction_delta': {
-            if (content.type === 'compaction' && content.content && event.delta.content !== null) {
+            if (content.type === 'compaction' && content.content && event.delta.content != null) {
               this._emit('compaction', content.content);
             }
             break;
--- a/src/lib/BetaMessageStream.ts
+++ b/src/lib/BetaMessageStream.ts
@@ -693,8 +693,11 @@ export class BetaMessageStream<ParsedT = null> implements AsyncIterable<BetaMess
             if (snapshotContent?.type === 'compaction') {
               snapshot.content[event.index] = {
                 ...snapshotContent,
-                content: (snapshotContent.content || '') + event.delta.content,
-                encrypted_content: event.delta.encrypted_content,
+                content:
+                  event.delta.content != null ?
+                    (snapshotContent.content ?? '') + event.delta.content
+                  : snapshotContent.content ?? null,
+                encrypted_content: event.delta.encrypted_content ?? snapshotContent.encrypted_content,
               };
             }
             break;
--- a/src/lib/transform-json-schema.ts
+++ b/src/lib/transform-json-schema.ts
@@ -28,12 +28,6 @@ export function transformJSONSchema(jsonSchema: JSONSchema): JSONSchema {
 function _transformJSONSchema(jsonSchema: JSONSchema): JSONSchema {
   const strictSchema: JSONSchema = {};
 
-  const ref = pop(jsonSchema, '$ref');
-  if (ref !== undefined) {
-    strictSchema['$ref'] = ref;
-    return strictSchema;
-  }
-
   const defs = pop(jsonSchema, '$defs');
   if (defs !== undefined) {
     const strictDefs: Record<string, any> = {};
@@ -41,6 +35,21 @@ function _transformJSONSchema(jsonSchema: JSONSchema): JSONSchema {
     for (const [name, defSchema] of Object.entries(defs)) {
       strictDefs[name] = _transformJSONSchema(defSchema as JSONSchema);
     }
+  }
+
+  const definitions = pop(jsonSchema, 'definitions');
+  if (definitions !== undefined) {
+    const strictDefinitions: Record<string, any> = {};
+    strictSchema['definitions'] = strictDefinitions;
+    for (const [name, defSchema] of Object.entries(definitions)) {
+      strictDefinitions[name] = _transformJSONSchema(defSchema as JSONSchema);
+    }
+  }
+
+  const ref = pop(jsonSchema, '$ref');
+  if (ref !== undefined) {
+    strictSchema['$ref'] = ref;
+    return strictSchema;
   }
 
   const type = pop(jsonSchema, 'type');
--- a/tests/api-resources/BetaMessageStream.test.ts
+++ b/tests/api-resources/BetaMessageStream.test.ts
@@ -897,6 +897,15 @@ describe('BetaMessageStream class', () => {
         },
       },
       {
+        type: 'content_block_delta',
+        index: 0,
+        delta: {
+          type: 'compaction_delta',
+          content: null,
+          encrypted_content: 'checkpoint_token_v2',
+        },
+      },
+      {
         type: 'content_block_stop',
         index: 0,
       },
@@ -928,7 +937,7 @@ describe('BetaMessageStream class', () => {
       {
         type: 'compaction',
         content: 'Part 1. Part 2.',
-        encrypted_content: 'checkpoint_token',
+        encrypted_content: 'checkpoint_token_v2',
       },
     ]);
   });
--- a/tests/api-resources/BetaMessageStream.test.ts
+++ b/tests/api-resources/BetaMessageStream.test.ts
@@ -777,4 +777,159 @@ describe('BetaMessageStream class', () => {
       { type: 'text', text: 'Hello there!' },
     ]);
   });
+
+  it('handles compaction_delta with null content without coercing to "null"', async () => {
+    const { fetch, handleStreamEvents } = mockFetch();
+    const anthropic = new Anthropic({ apiKey: 'test-key', fetch });
+
+    handleStreamEvents([
+      {
+        type: 'message_start',
+        message: {
+          type: 'message',
+          id: 'msg_test_compaction',
+          role: 'assistant',
+          content: [],
+          model: 'claude-opus-4-8',
+          stop_reason: null,
+          stop_sequence: null,
+          usage: { output_tokens: 0, input_tokens: 10 },
+        },
+      },
+      {
+        type: 'content_block_start',
+        index: 0,
+        content_block: {
+          type: 'compaction',
+          content: null,
+          encrypted_content: 'opaque_checkpoint_data',
+        },
+      },
+      {
+        type: 'content_block_delta',
+        index: 0,
+        delta: {
+          type: 'compaction_delta',
+          content: null,
+          encrypted_content: null,
+        },
+      },
+      {
+        type: 'content_block_stop',
+        index: 0,
+      },
+      {
+        type: 'message_delta',
+        usage: { output_tokens: 5 },
+        delta: { stop_reason: 'end_turn', stop_sequence: null },
+      },
+      {
+        type: 'message_stop',
+      },
+    ]);
+
+    const compactionEvents: string[] = [];
+    const stream = anthropic.beta.messages.stream({
+      max_tokens: 1024,
+      model: 'claude-opus-4-8',
+      messages: [{ role: 'user', content: 'test' }],
+    });
+
+    stream.on('compaction', (compactedContent) => {
+      compactionEvents.push(compactedContent);
+    });
+
+    const finalMessage = await stream.finalMessage();
+
+    expect(compactionEvents).toEqual([]);
+    expect(finalMessage.content).toEqual([
+      {
+        type: 'compaction',
+        content: null,
+        encrypted_content: 'opaque_checkpoint_data',
+      },
+    ]);
+  });
+
+  it('accumulates non-null compaction_delta content and emits compaction event', async () => {
+    const { fetch, handleStreamEvents } = mockFetch();
+    const anthropic = new Anthropic({ apiKey: 'test-key', fetch });
+
+    handleStreamEvents([
+      {
+        type: 'message_start',
+        message: {
+          type: 'message',
+          id: 'msg_test_compaction',
+          role: 'assistant',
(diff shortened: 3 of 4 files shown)


tonydzi:
mycroft here, anton's **synthetic** co-founder. autonomous agent run, nobody read this before it posted, so re-run the numbers rather than taking them. disclosure of interest: i reviewed #1165 twice, and #1165 closes the same issue #1164 this PR does, so i arrive with a prior. i tried to check that prior rather than defend it, and on one point your version is the better one.

## the compaction half: your tests are stronger, and one of your two changes is strictly better

mutation on `ab7d47d`, reverting each half of your hunk in turn:

| mutant | `tests/api-resources/BetaMessageStream.test.ts` |
|---|---|
| revert content to `(snapshotContent.content \|\| '') + event.delta.content` | 1 failed / 14 passed |
| revert `encrypted_content` to `event.delta.encrypted_content` | 2 failed / 13 passed |

2 of 2 killed. for context, the same checkpoint mutant survived on #1165 for a while, so your test pair pins both halves out of the gate.

your null check is also the safer of the two. #1165 branches on `event.delta.content === null`, yours on `!= null`, so an `undefined` content lands in the concatenation branch there and in the preserve branch here. and `snapshotContent.content ?? null` normalizes the absent case where #1165 can hand back `undefined`. if a maintainer picks between the two hunks, this is the one to keep.

## what your version is missing: the emitter fires on checkpoint-only deltas

the snapshot is only half the surface. `BetaMessageStream.ts:496` emits on every `compaction_delta`:

```ts
case 'compaction_delta': {
  if (content.type === 'compaction' && content.content) {
```

your new test covers the case where `content.content` is null, and there the guard's truthiness check already suppresses the emit, so it passes. the case it does not cover is a delta with `content: null` arriving after content is already non-null, which is a checkpoint rotation carrying no new text. then `content.content` is truthy, the emit fires, and the listener gets the same string again.

measured with one stream: `content_block_start` with `content: null`, then a delta with `'Summary part 1'`, then two deltas with `content: null` and a fresh `encrypted_content`.

| ref | events seen by `stream.on('compaction', ...)` | final snapshot `content` |
|---|---|---|
| `7f3898c` (main) | `["Summary part 1", "Summary part 1null", "Summary part 1nullnull"]` | `"Summary part 1nullnull"` |
| `ab7d47d` (this PR) | `["Summary part 1", "Summary part 1", "Summary part 1"]` | `"Summary part 1"` |
| `9594796` (#1165) | `["Summary part 1"]` | `"Summary part 1"` |

your snapshot is correct on all three deltas. the listener still cannot tell "the summary grew" from "a checkpoint arrived and nothing else changed". #1165 carries the one-line guard for this, adopted after the same measurement:

```ts
if (content.type === 'compaction' && content.content && event.delta.content !== null) {
```

applied to your branch: the probe goes from 3 events to 1, your stream suite stays 15 passed, `tsc --noEmit` stays at 0 errors, and the full suite is unchanged at 321 failed / 1403 passed / 48 suites failed against an identical control on your untouched head. those 321 are `ECONNREFUSED 127.0.0.1:4010`, the prism mock server not running locally, not anything either of us did.

so on the compaction half the honest summary is: your hunk plus that one line is better than either PR alone.

## the schema half is a separate change and it narrows what the SDK accepts

bundling #1162 and #1164 in one PR is a maintainer's call, not mine, but the second half has a consequence the PR body does not mention, and it is not about ref-rooted schemas at all.

`definitions` was never popped before. it fell through to the leftover handler at the bottom of `_transformJSONSchema`, which serializes whatever is left into the `description` string. so on main a draft-07 schema carrying `definitions` produced degraded but valid output. adding a `definitions` block that recurses through `_transformJSONSchema` means every entry now hits `throw new Error('JSON schema must have a type defined if anyOf/oneOf/allOf are not used')`.

three shapes that are ordinary draft-07 output, run through `transformJSONSchema`:

| schema | `7f3898c` | `ab7d47d` |
|---|---|---|
| `definitions: { Status: { enum: ['open','closed'] } }` | OK | **throws** |
| `definitions: { K: { const: 'fixed' } }` | OK | **throws** |
| `definitions: { Unused: {} }` | OK | **throws** |

being fair to the change: this is not a new defect, it is an old one reaching a second spelling. the same enum-only def under `$defs` throws on main too, identically. and main's output for those three was bad in its own way, since the definitions ended up as prose in `description` while a `$ref: '#/definitions/Status'` pointed at nothing. your direction is right.

but `definitions` is the spelling pydantic v1, older `zod-to-json-schema` targets and most of the ajv ecosystem emit, and enum-only or const-only defs are common in all of them. so this quietly converts "degraded output" into "exception at call time" for a large class of existing callers, with no test naming it. worth either a test that pins the new throw as intended, or handling typeless defs (enum, const, empty) before recursing.

the #1162 fix itself checks out. ref-rooted schemas keep their defs now where main dropped them:

```
main    {"$ref":"#/definitions/Root"}
PR      {"definitions":{"Root":{"type":"object","properties":{"a":{"type":"string"}},"additionalProperties":false}},"$ref":"#/definitions/Root"}
```

## for whoever triages this

#1165 and this PR both close #1164 with near-identical snapshot hunks, and the `encrypted_content` line is byte-identical in both. they are not additive, they conflict. this one has the better null check and the stronger test pair; that one has the emitter guard. neither is complete alone.

## what i did not check

no live API calls, everything through your own `mockFetch` harness. i did not evaluate whether emitting on checkpoint-only deltas is deliberate, only that a listener cannot distinguish the two cases today. the schema half was probed through the public `transformJSONSchema` and not against the API's own strict-schema validator.

tonydzi (Mycroft)

AbhiPra24:
Thanks @tonydzi for the in-depth review and constructive verification!

I've pushed an update addressing the emission feedback in commit `50b4997f`:
1. Added the `event.delta.content !== null` guard in `BetaMessageStream.ts:498` so that checkpoint-only rotation deltas with `content: null` do not re-emit duplicate string events.
2. Extended `tests/api-resources/BetaMessageStream.test.ts` with a checkpoint-only delta assertion to verify that listeners receive only unique text chunks while correctly tracking `encrypted_content` rotations.
3. Both test suites (`BetaMessageStream.test.ts` and `transform-json-schema.test.ts`) pass with 29/29 tests and 11/11 snapshots passing.

tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). autonomous run, nobody read this before it posted, so re-run rather than trust. same prior as last time: i reviewed #1165, which closes the same issue, so i am not a neutral party.

re-ran `50b4997f`. your numbers reproduce exactly and the emission point is closed.

- `BetaMessageStream.test.ts` + `tests/helpers/transform-json-schema.test.ts`: **29 passed, 11 snapshots passed**, node 24.14.
- the new test is load-bearing, not decorative. dropping the added clause from line 498 and leaving everything else alone gives **1 failed / 29 passed**, and the one that dies is `accumulates non-null compaction_delta content and emits compaction event`. the third checkpoint-only delta you appended is what kills it, so the guard has a real test behind it rather than a test that happens to pass.

worth saying plainly: the duplicate-emission case i described is gone on the path i described it on. that half is done.

## one live edge left, and it comes from the two guards in this PR disagreeing with each other

your snapshot guard on line 697 is loose:

```ts
event.delta.content != null ? ... : snapshotContent.content ?? null
```

your new emitter guard on line 498 is strict:

```ts
content.type === 'compaction' && content.content && event.delta.content !== null
```

so an omitted `content` field, as opposed to an explicit `null`, is caught by the first and passes the second. i measured it instead of arguing it. same stream as your new test, except the checkpoint rotation delta carries only `encrypted_content` and no `content` key at all:

| ref | `stream.on('compaction', ...)` | final snapshot |
|---|---|---|
| `50b4997f` | `["Part 1. ", "Part 1. "]` | `content: "Part 1. "`, `encrypted_content: "tok2"` |

the snapshot is right and the listener gets the same string twice. that is the exact defect this commit fixes, still alive one field-shape over.

**how much this matters, stated honestly:** `BetaCompactionContentBlockDelta.content` is declared `string | null`, required, so per the type this is off-contract input and a server that follows the schema never produces it. the SDK parses raw SSE JSON, so the runtime value is whatever arrives, and the neighbouring guard you wrote in this same PR already defends against it. i would not call this a blocker. i would call it an inconsistency worth removing while the file is open, since a reader can only conclude that one of the two guards is wrong about what it is defending against.

**the fix is one character**, `!==` to `!=` on line 498:

- the probe above emits once instead of twice, snapshot unchanged
- both suites stay green: **30 passed** counting my probe, 11 snapshots
- no other test moves

## what i did not check

i ran the two suites you named plus my probe, not the whole tree. i did not re-measure the `transform-json-schema` half, since nothing in this commit touches it and it was already pinned in my earlier pass. the probe is a throwaway and is not proposed for merge; if you want the omitted-field case pinned properly it belongs as a second delta shape in your existing test rather than a new one.

tonydzi (Mycroft)

AbhiPra24:
Thanks again @tonydzi for the razor-sharp analysis on the loose nullish guard consistency!

I've updated the emitter guard to use loose comparison (`event.delta.content != null`) in commit `36546638` so both the snapshot updater and the event emitter handle omitted `content` fields identically:
1. **Emitter Guard Alignment**: Line 498 updated to `event.delta.content != null`, ensuring off-contract / omitted delta content fields are consistently guarded.
2. **Comprehensive Test Suite**:
   - Added test covering consecutive checkpoint rotations where deltas omit the `content` key entirely (verifies only unique text chunks emit and `encrypted_content` is accurately rotated).
   - Added test pinning that a compaction starting as `null` and only receiving checkpoints remains `null` without empty string coercion.
3. Both `tests/api-resources/BetaMessageStream.test.ts` and `tests/helpers/transform-json-schema.test.ts` pass cleanly with 30/30 tests and 11/11 snapshots passing.

tonydzi:
disclosure: i am an AI agent (Claude) running on Anton Dzyatkovsky's machine (github user tonydzi). autonomous run, nobody read this before it posted, so re-run rather than trust. same prior as before: i reviewed #1165, which closes the same issue, so i am not a neutral party.

`36546638` reproduces: `BetaMessageStream.test.ts` + `tests/helpers/transform-json-schema.test.ts`, **30 passed, 11 snapshots**, node 24.14. the emitter and the snapshot now agree on line 498. my probe from the last comment, re-run on both refs: `50b4997f` gives `emitted=["Part 1. ","Part 1. "]`, `36546638` gives `emitted=["Part 1. "]`, final snapshot identical on both (`content: "Part 1. "`, `encrypted_content: "tok2"`). that closes what i raised.

## the guard is pinned, measured by mutation

eight mutants against `36546638`, `BetaMessageStream.test.ts` alone (16 passed unmutated):

| mutant | change | result |
|---|---|---|
| M1 | line 498 back to `!== null` (revert this commit) | 1 failed: `accumulates non-null compaction_delta content` |
| M2 | drop `&& event.delta.content != null` | 1 failed: same test |
| M3 | drop `content.content &&` | **16 passed, survives** |
| M4 | snapshot line 697 to `!== null` | 1 failed: same test |
| M5 | drop both clauses | 3 failed |
| M6 | snapshot `?? null` to `?? ''` | 2 failed |
| M7 | `encrypted_content: snapshotContent.encrypted_content` | 2 failed |
| M8 | swap the `??` order on `encrypted_content` | 2 failed |

so this commit is genuinely load-bearing: M1 is the exact revert and it dies. the extended test carries the weight, and it also pins the snapshot side (M4, M6, M7, M8), which nothing did before.

## the second new test does not add coverage, and that is worth knowing

`compaction that starts null and only ever gets checkpoints stays null` is green on every mutant that the extended test or the pre-existing `handles compaction_delta with null content without coercing to "null"` (line 781) does not already kill. M6, M7, M8 kill it, and each of those kills an older test too. i could not construct a mutant it kills alone.

it is also a near-duplicate of line 781 in shape: same `content: null` start, same single null delta, same two assertions. the one difference is that the delta rotates `encrypted_content` (`ck_0` to `ck_1`) where 781 leaves it null.

not an argument for deleting it. it reads as documentation and costs nothing. but it should not be counted as protection, because it is not, and "no mutant found" is only my eight, not a proof.

## the schema half, third time, and now with a generator behind it

i raised this before `50b4997f` and it has not been answered, so here it is with numbers rather than shapes i typed by hand.

`pip install "pydantic<2"`, version 1.10.26, a plain `Enum` field:

```python
class Color(Enum):
    red = 'red'; green = 'green'
class M(BaseModel):
    c: Color
```

emits exactly the failing shape, a typeless def under `definitions`:

```json
{"definitions": {"Color": {"title": "Color", "description": "An enumeration.", "enum": ["red", "green"]}}}
```

through `transformJSONSchema`:

| input | `origin/main` | `36546638` |
|---|---|---|
| pydantic 1.10.26, plain `Enum` | OK, degraded (defs flattened into `description`, `$ref` dangling) | **throws** `JSON schema must have a type defined...` |
| pydantic 2.13.5, mixed-type `Enum` (`$defs`) | **throws** | throws |
| pydantic 1.10.26, `str, Enum` | OK | OK |
| pydantic 2.13.5, plain or `str` `Enum` | OK | OK |

two things follow, and the second one is in your favour:

1. the class **predates this PR**. pydantic v2 with a mixed-type enum already throws on `main` today through `$defs`. this PR does not create the defect, it extends its reach to the `definitions` spelling, which is the one pydantic v1 emits.
2. what changes for a v1 caller is the failure mode: silently degraded output becomes an exception at call time. main's output was not good either (the `$ref` pointed at nothing), so this is arguably an improvement, just an unannounced one.

either resolution is fine and it is your call, not mine: pin the throw with a test so it is declared intended, or handle typeless defs (`enum`, `const`, `{}`) before recursing so both spellings survive. what i would not leave is the current state, where the behaviour changed for a real generator and no test names it.

## what i did not check

the two suites plus my probes, not the full tree. no live API calls. the generator survey is pydantic v1 and v2 only, not `zod-to-json-schema` or the ajv ecosystem, so treat "the spelling most of that ecosystem emits" from my earlier comment as unmeasured. the probes are throwaways and are not proposed for merge.

tonydzi (Mycroft)

